PR #18, opened 2020-09-29
Restricts to exact versions

arcanis:
Ranges are a tad dangerous because there's no telling which version will be installed exactly. While they are supposed to be compatible, the pros and cons are somewhat unbalanced at the moment: the risk of a regression is higher than the potential size gain, because Corepack doesn't cover npm at the moment - which is the main case where deduplication would matter.

It's better to restrict the `packageManager` field to strict semver versions for this first iteration and see if there's a need for ranges (which would be possible to add in a minor) than do the opposite and have to make a backward-incompatible change later down the road.

### DIFF
--- a/package.json
+++ b/package.json
@@ -8,7 +8,7 @@
     "yarn": "./dist/yarn.js",
     "yarnpkg": "./dist/yarnpkg.js"
   },
-  "packageManager": "yarn@^2.0.0-rc.29",
+  "packageManager": "yarn@2.0.0-rc.29",
   "devDependencies": {
     "@babel/core": "^7.11.0",
     "@babel/plugin-proposal-class-properties": "^7.10.4",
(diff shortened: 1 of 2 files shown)


TemaSM:
How about to add such info into Readme?
I mean, it was unexpected, that `package.json` should have only semver-compatible version, but not range of versions. And even [Nodejs docs](https://nodejs.org/api/packages.html#packagemanager) lacks of explicitly stated info about versions ranges or strict mode inside `packageManager` field.

sodatea:
I find it really counter-intuitive that we have an exact version in `package.json`. It's usually the job of lockfiles, and can be auto-updated once we run a successful `yarn` / `pnpm install` / `npm install`.

How about adding a `packageManager` field in each package manager's lockfile spec and letting corepack read that instead?

mail731212 (commented):
Kun731020[]()

PR #19, opened 2020-09-30
Implements transparent commands

### LINKED ISSUE #1: Global packages
This thread aims to discuss the "global packages" scenario. For example, the pmm version of `pnpx --version` is currently forbidden if run within a Yarn project, even though it's supposed to run within a standalone context that might not be related.

### LINKED ISSUE #2: Init workflow
This thread is meant to discuss the init workflow. For example, what happens when running `yarn init` or `npm init`? Under the current model those commands will fail, because pmm is implemented in such a way that the `package.json` is created before spawning the process, which will then detect the preexisting manifest and refuse to update it.

arcanis:
It's currently fairly annoying to run the `init` commands, because they go through the jumper, which detects that there is no local project, asks if it should create one, spawns the init command, and fails because there's already a package.json. Additionally, running `yarn dlx` or `pnpx` isn't possible at the moment from (respectively) non-Yarn or non-pnpm projects.

This diff fixes that by defining a set of "transparent commands". The jumpers will allow those commands to be used regardless of the general environment - making it possible to run `pnpx` within Yarn projects, and running `yarn init` without prompts.

Fixes #1, fixes #2


arcanis:
I've also removed Enquirer. While it's a good idea in theory, it's very impractical as it:

- causes spurious prompts to appear when running in the wrong directories
- causes `package.json` to appear on the filesystem - problematic if you're for example in the home folder

Instead, Corepack will silently accept any package manager and use their LKG releases (similar to what already happens when you're inside a non-configured project). It doesn't change much, the only downside being that the `packageManager` field adoption will be lower, but it's worth it if it means less disruption for the users. Additionally, the `init` commands (such as `yarn init`) still have the ability to set the `packageManager` field themselves if they wish to.

PR #20, opened 2020-09-30
fix: handle package managers with a bin array correctly

merceyz:

In https://github.com/arcanis/corepack/pull/13 I broke running the `yarnpkg` command because I didn't handle the array correctly

**How did you fix it?**

Handle package managers with an "array bin field" correctly, by doing the same parsing in `pmmUtils.runVersion` as in `pmmUtils.installVersion`


Uzomahopeujunwa (commented):
Lovely

Issue #25, opened 2020-10-30
Package manager detection


arcanis:
We were discussing with @iansu the case where package managers are detected depending on whether they are or not in the global path (@ljharb mentioned that as well in another thread). Tracking some thoughts:

- Existing CRA & similar tutorials mention npm, so newcomers might be surprised if it gets installed via Yarn.

- At the same time, it would be similarly confusing if tools like CRA were to always default to npm.

- If we do this Corepack thing, it's at least in part so that all package managers are on the same foot.

- The problem affects scaffolders, since existing projects will necessarily already have their install creations.

One potential solution would be to offer a [BrowserChoice](https://en.wikipedia.org/wiki/BrowserChoice.eu)-like system under the form of a Corepack command (let's say `corepack elect` for now) that, when called:

- If the command was never run before, it would print all supported package managers with an interactive interface listing their name and a link towards their respective websites. Upon selecting one of the options, Corepack would save the result inside its data folder, then would print the name of the selected package manager.

- If the command ran before, it would simply read the data previously stored inside the data folder, and print again without prompting the user.

From the integrator perspective, the way it would work:

- Instead of checking for Yarn's existence with things like `yarn --version`, the program would run `corepack elect` and capture the output. It would then branch depending on the result as before.

- They could still implement their flags like `--use-npm` or `--use-yarn`, so that users can override their choice on a per-command basis.

One nice effect is that this would make it much clearer for the users which package manager they would be using. Right now, it's difficult to know whether Yarn or npm will end up being used unless 1/ you know that CRA has this behavior, and 2/ you remember whether you installed Yarn or not (you may have installed it only for another project a long time ago).

On the minus side it means that tools like CRA would have to add support for this feature, so it would require their input to know how they would feel about this workflow.

iansu:
I like this idea in general, both as a user and as a tooling author. cc @nodejs/tooling 

The one concern I have from a Create React App perspective is that our whole thing is that we're a zero-config tool. We go out of our way to avoid prompting the user for anything ever.

The biggest reason for this is that a lot of our users are beginners, new to React, new to JavaScript and sometimes even new to programming. Every prompt is an obstacle to the user. If they don't even know what a package manager is or have never heard of npm or yarn or pnpm they might get stuck.

Aside from that, the idea of being able to programatically ask for the user's preferred package manager and use that by default is very appealing.

arcanis:
I've implemented a first draft in #26

nickmccurdy:
[...]

PR #26, opened 2020-11-23
Implements "corepack elect"

arcanis:
See #25 and the [documentation](https://github.com/nodejs/corepack/blob/mael/elect/README.md#elected-package-manager) added to `README.md` for more details.

In short, this PR implements a proper way for projects such as create-react-app to detect which package manager to use when initializing a project, by gently asking the user only once for all installs that will occur on the machine.


BertsLifee (approved):

merceyz (changes requested):
I'm -1 on this, I would prefer if packages check `process.env.npm_config_user_agent` to see which package manager was used and then do whatever they need to based on that.

BertsLifee (approved):

styfle:
I agree with merceyz.

For example, `create-next-app` [documents these options](https://nextjs.org/docs/pages/api-reference/create-next-app) for installation:

```sh
npx create-next-app@latest
yarn create next-app
pnpm create next-app
```

And then it uses [`process.env.npm_config_user_agent`](https://github.com/vercel/next.js/blob/d6a3e7d07b6e9f2622463bfbf3a49555173fd5cb/packages/create-next-app/helpers/get-pkg-manager.ts#L4) to determine which package manager was used.

arcanis:
Yeah, it's an old experiment, from prior those options were implemented. I don't plan to land it anymore.

PR #34, opened 2021-05-17
docs: fix pmm -> corepack in Contributing

shisama:
The `pmm` was refactored to `corepack` (9809b8f856cba5e91a1ec54f595073373696440d). However, `pmm` has still been used in README. I think it is better to change it. Also, `dist/corepack` does not have the `--version` option, so I fixed it with `--help`.


AngelicaOvesdotter:
image

AngelicaOvesdotter:
image

PR #36, opened 2021-05-31
Modernize the dependencies

arcanis:
- Bumps dependencies (most notably Clipanion, TypeScript, Yarn, Jest, Babel, Webpack)
- Adds v8-compile-cache when spawning the package managers
- Changes the Yarn 2 download URL to reference our website (cached behind Cloudflare), to avoid GH's rate limiting
- Removes the git tag fetching to replace it by straight https queries (less reliance on GH's infra)
- Optimizes the happy path boot time by moving `tar` and `https` into lazy imports


package.json
```diff
@@ -19,14 +19,14 @@
   },
   "packageManager": "yarn@2.0.0-rc.29",
```
merceyz, line 20: Could you update this version as well please, otherwise corepack will spawn an old version of v2 which fails on parsing newer settings in `~/.yarnrc.yml`

merceyz (commented):
Can't make it a suggestion but we should update the version at https://github.com/nodejs/corepack/blob/2a46b6d13adae139141012254ef670d6ddcb5d11/config.json#L50 as well

package.json
```diff
@@ -87,5 +87,9 @@
       "./shims/yarnpkg",
       "./shims/yarnpkg.ps1"
     ]
+  },
+  "dependencies": {
+    "babel-plugin-dynamic-import-node": "^2.3.3",
+    "v8-compile-cache": "^2.3.0"
```
merceyz, line 93: These should probably be in `devDependencies`

arcanis: Indeed! v8-compile-cache originally was a runtime dep, but it proved impractical so I bundled it

PR #42, opened 2021-07-31
Fixes the displayed version number

arcanis:
The version wasn't properly set, so running `corepack -v` didn't work.


aduh95:
This requires `corepack` to be available globally:

```console
$ sh ../../node/deps/corepack/shims/corepack -v                           
Error: Cannot find module 'corepack/package.json
```

PR #58, opened 2021-09-29
Bump Yarn to 1.22.15

### LINKED ISSUE #44: Access to the binaries without global install
I wonder if it would be possible to have a feature similar to `python -m pip` to access the package manager binaries without having them on the path. It would be useful in case the binary name clashes with another program. That could potentially address the concern of "How Corepack scales if there's 1000 package managers added to it, each package manager adding 100 binaries to the path?".

The API could look something like:

```console
> corepack yarn install
```

wdyt?

arcanis:
Fixes `yarn init -2`, which broke due to the changes that fixed #44


arcanis:
I'll have to find a better solution for the network mocks, they are completely unstable 😤

PR #69, opened 2021-11-19
feat: auto setup proxy for http requests

### LINKED ISSUE #67: Install package manager behind a corporative proxy
I tried to do, as instance `corepack prepare pnpm@5.18.10`, but got `Internal Error: Error when performing the request`.
I think this is because of the proxy. Where I can configure it? 
I have proxy setting in `.npmrc`, also I have `HTTP_PROXY` and `HTTPS_PROXY` environment variables

Djaler:
Resolves #67


sources/httpUtils.ts
```diff
@@ -23,6 +25,19 @@ export async function fetchUrlStream(url: string, options: RequestOptions = {})
   });
 }
 
+let globalAgentBootstrapped = false;
+
+async function setupProxy() {
+  if (globalAgentBootstrapped) {
```
Djaler, line 31: Maybe this is not required if we know for sure that corepack will execute only one request in process lifetime.

Djaler:
@arcanis so what about it?

Djaler:
I don't want to annoy, but I really need this feature, because for now we can't use corepack on work computers

arcanis:
Hey @Djaler, apologies for the delay, I should have given feedback sooner 🙁

While `global-agent` does work, I'd tend to prefer to avoid global mutation of the environment (at the moment we use `spawn` to spawn the underlying package managers, but it could be conceivable to use `require` instead to avoid part of the boot overhead). Would there be any chance we could do this by having a custom agent (created and local to `httpUtils`) that we could manually pass to `https.get`?

Djaler:
@arcanis sorry for a big delay too.

Looks like I found a good way to do this without global mutations, yeah

Djaler:
About type error - I don't know why it fails. `proxy-agent` provides typings https://github.com/TooTallNate/node-proxy-agent/blob/master/index.d.ts

julioz:
I'm sorry to be "that person" but like @Djaler, the lack of support for https proxying is what prevents us from using Corepack as well. @arcanis @Djaler is there something I can do to help move this PR forward?

Djaler:
Hey, @arcanis so what about it?

arcanis:
LGTM, thanks @Djaler !

Djaler:
@arcanis any chance to release this feature soon?

PR #78, opened 2021-12-31
Updated readme looks on packageManager Usage

frog-o:
It was hard for me to find this info in your file.  I clean it up so I could find it better.

### DIFF
--- a/README.md
+++ b/README.md
@@ -36,18 +36,36 @@ We do acknowledge the irony and overhead of using npm to install Corepack, which
 
 ## Usage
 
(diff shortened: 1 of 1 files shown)


arcanis:
While documenting things that aren't already documented is always fine, rewording things to make them clearer also has the risk to make them less clear to someone else, so I'm not sure this PR is mergeable.

frog-o:
One thing i hate is getting rid of of contact and rewording things for just
the sake of rewording. If you read closely I did not change much. The
wording perty much the same with a more comprehensible layout at least in
my case since ( I have a disability) a really big problem with
comprehension; please keep this in mind.  I would never ask you to change
something that makes it unreadable to you though.

arcanis:
I'll close this PR as I'm still unconvinced those changes are for the better.

PR #82, opened 2022-01-23
ci: add cache validation

merceyz:

Some of the entires in `.yarn/cache` has the wrong checksum and there isn't a CI check to prevent that from happening.

**How did you fix it?**

Updated the checksums and added cache validation to the CI.


arcanis (approved):
Just checked the versions currently in cache, the only changes are the timestamp ones when we switched Yarn's normalized time from 1980 to 1984; files have the same content before and after 👍

PR #84, opened 2022-01-23
fix: handle parallel installs

merceyz:

When multiple instances of Corepack try to install the same version of a package manager one of them will fail with an `ENOTEMPTY` error.

**How did you fix it?**

Catch the error and assume another instance of Corepack installed the version first and clean up the temp folder


sources/fsUtils.ts
```diff
@@ -1,3 +1,24 @@
...
+    const rimraf = await import(`rimraf`);
+    return new Promise<void>((resolve, reject) => {
+      rimraf.default(path, err => {
+        if (err) {
+          reject(err);
+        } else {
+          resolve();
+        }
+      });
+    });
```
arcanis, line 22: We can remove this code in April

arcanis (approved):

PR #94, opened 2022-02-25
tools: add `coverage` folder to `.gitignore`

austinkelleher:
When running `jest --collect-coverage` a `coverage` folder is created,
which should not be committed.


austinkelleher:
@arcanis FYI. I plan is to add some additional tests to corepack soon.

PR #99, opened 2022-04-16
Fixes `corepack prepare` when no additional arguments are passed

### LINKED ISSUE #96: 'corepack prepare' does not read "packageManager" in package.json
The [docs](https://github.com/nodejs/corepack#corepack-prepare--nameversion) and [comments](https://github.com/nodejs/corepack/blob/1d41c484800dc043d0c713aba6a4e3bbc0e1ee09/sources/commands/Prepare.ts#L22-L23) indicate that `corepack prepare` should look at the package manager configured in `package.json` and download it, but it currently doesn't.

## Repro
1. `docker run --rm -it --entrypoint sh node:16.11.1-alpine3.14`
2. `echo '{"packageManager": "yarn@3.0.2"}' > package.json`
3. `ls ~/.node` doesn't exist (expected)
4. `corepack prepare`
5. `ls ~/.node` doesn't exist **(unexpected)**
6. `corepack yarn --version` (shows 3.0.2)
7. `ls ~/.node` does exist (expected)

kherock:
Fixes #96

The existing implementation of `prepare` iterates over an array of `string | Descriptor` requests and only looks up the project version when `typeof request === 'undefined'`, which is never true. This wasn't caught by the test suite since the network was left enabled.

I've updated the command to instead look up the project spec when the specs list is empty, before the loop starts.

### DIFF
--- a/.yarn/sdks/typescript/lib/tsserver.js
+++ b/.yarn/sdks/typescript/lib/tsserver.js
@@ -18,6 +18,7 @@ const moduleWrapper = tsserver => {
   const pnpApi = require(`pnpapi`);
 
   const isVirtual = str => str.match(/\/(\$\$virtual|__virtual__)\//);
+  const isPortal = str => str.startsWith("portal:/");
   const normalize = str => str.replace(/\\/g, `/`).replace(/^\/?/, `/`);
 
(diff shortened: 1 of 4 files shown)


arcanis:
LGTM, thanks @kherock!

PR #100, opened 2022-04-17
chore: update dependencies, remove unused ones

aduh95:
- This PR removes some dependencies that are no longer necessary as long as we're OK with dropping v12.x support (which turns EOL on April 30th).
- To update the Yarn version used to manage Corepack dependencies, this repo is now using Corepack, so you now need Corepack to build Corepack (but imo that's OK now that Corepack is bundled with all supported Node.js versions).
- I've updated all outdated npm dependencies and GitHub actions.


merceyz (changes requested):
The `prepack` script wont work anymore since the file it points to was removed.

We should add a `engines` property to `package.json` to actually document the supported Node versions.

arcanis:
The Windows tests are quite flaky, since they rely on the network 😕

aduh95:
> The Windows tests are quite flaky, since they rely on the network 😕

It's probably related to the changes in this PR though, as the flakyness seems to be quite consistent, but I haven't a Windows machine available to test that 😕

aduh95:
I've just tested it on a Windows machine, and I'm not able to reproduce the failure (tested with Node.js 16.15.0 and Node.js 18.0.0), so it is probably a fluke indeed. Let's wait for Node.js 14.19.2 to be released next week as it will contain the fix for the Jest issue on Node.js 14.x.

merceyz:
> The Windows tests are quite flaky, since they rely on the network 😕

Isn't `nock` supposed to make it not rely on the network?

aduh95:
It looks like the test flakiness doesn't exist on Node.js 14.x, tests are now green :)

.github/workflows/ci.yml
```diff
@@ -8,57 +8,57 @@ on:
...
-      - uses: actions/checkout@v2
+      - uses: actions/checkout@v3
 
       - run: |
           git fetch --no-tags --unshallow origin HEAD main
 
-      - uses: actions/setup-node@v2
+      - uses: actions/setup-node@v3
         with:
-          node-version: 14
```
arcanis, line 22: Is something broken in the latest Node 14 that requires pinning?

.github/workflows/ci.yml
```diff
@@ -8,57 +8,57 @@ on:
...
         platform:
-        - ubuntu-latest
-        - macos-latest
-        - windows-latest
+          - ubuntu-latest
+          - macos-latest
+          - windows-latest
 
-    name: '${{matrix.platform}} w/ Node.js ${{matrix.node}}.x'
+    name: "${{matrix.platform}} w/ Node.js 14.x"
```
arcanis, line 49: Imo we should keep the `matrix.node` variable here (just remove the `.x`)
```suggestion
    name: "${{matrix.platform}} w/ Node.js ${{matrix.node}}"
```

.github/workflows/build.yml
```diff
@@ -8,38 +8,38 @@ jobs:
...
-          git commit -m 'Updates corepack'
-          git show --name-status
-        fi
-        ./configure
-        DISTTYPE=nightly DATESTRING=YYYY-MM-DD COMMIT=XXXX make binary -j8
-        mv node-v15.0.0-nightlyYYYY-MM-DDXXXX-*.tar.gz node-corepack-${{matrix.platform[0]}}-x64.tar.gz
+      - name: "Build Node with corepack master"
+        run: |
+          corepack yarn install --immutable
+          corepack yarn pack
```
arcanis, line 27: Why do you call `corepack yarn` rather than just `yarn` (same in the other places)?

aduh95: To make it explicit that we want to use the Corepack version of Yarn, in case Yarn v1.x is installed on the runner. Also, if someone is using this file to know what command to use to run the tests, it's the most likely command to succeed.

.github/workflows/ci.yml
aduh95: Something is broken with the all Node.js 14 versions but the latest (https://github.com/yarnpkg/berry/issues/4157).
The GHA runner is still picking up 14.19.1 (because their cache is outdated), pinning the version ensures the correct version is installed. This can be reverted in a few weeks I hope.

.github/workflows/build.yml
arcanis: In that case, perhaps calling `corepack enable` (or perhaps `sudo corepack enable`) beforehand would be better - from a CLI user experience point of view, I'd prefer if Corepack didn't require any change, to guarantee reduced friction.

.github/workflows/ci.yml
arcanis: Sounds good, thanks 👍

.github/workflows/ci.yml
arcanis: Ah I see - makes sense then, combined with your other comment 👍

.github/workflows/build.yml
aduh95: I agree that might be desirable for CLI usage, I'm doubtful it's a valid point in a CI context. The tests already ensures that calling the `corepack` binary is optional once `corepack enable` has been run, right?

Adding a `corepack enable` step in CI seems "wasteful" to me, but if you feel strongly about this, that's fine by me. wdyt?

arcanis:
Thanks @aduh95! I'll look to make a release soon, I just need to upgrade the checked-in versions

PR #103, opened 2022-05-03
fix pnpm v7

### LINKED ISSUE #102: pnpx breaks in corepack+pnpm v7
https://github.com/pnpm/pnpm/releases/tag/v7.0.0

> pnpx is now just an alias of pnpm dlx.

> If you want to just execute the command of a dependency, run pnpm <cmd>. For instance, pnpm eslint.

> If you want to install and execute, use pnpm dlx <pkg name>.

Now I cannot run a pnpx command inside an npm configured project even though it installs the package in the global.

I need it because I need to run the following command:

> pnpx -y npm@latest-6 install

Otherwise, npm v7+ will upgrade the lock file which is unwanted for projects that I cannot control.

I cannot use npx either because npx will always use the installed version of npm v8 and ignore the `latest-6` I specified.

image

Jack-Works:
close #102

I don't know if it works though... Is it possible to "preview" the changes on my local machine?


Jack-Works:
Looks like the test fails on macos only...

arcanis:
Thanks @Jack-Works !

PR #106, opened 2022-05-15
Updates package manager versions

arcanis:
Upgrade package manager versions to their latest releases.

### DIFF
--- a/config.json
+++ b/config.json
@@ -28,7 +28,7 @@
       }
     },
     "pnpm": {
-      "default": "7.1.0",
+      "default": "7.1.5",
       "transparent": {
         "commands": [
           [
--- a/config.json
+++ b/config.json
@@ -1,7 +1,7 @@
 {
   "definitions": {
     "npm": {
-      "default": "8.4.0",
+      "default": "8.10.0",
       "transparent": {
         "commands": [
           [
@@ -28,7 +28,7 @@
       }
     },
     "pnpm": {
-      "default": "6.29.1",
+      "default": "7.1.0",
       "transparent": {
         "commands": [
           [
@@ -70,9 +70,9 @@
       }
     },
     "yarn": {
-      "default": "1.22.17",
+      "default": "1.22.18",
       "transparent": {
-        "default": "3.1.1",
+        "default": "3.2.1",
         "commands": [
           [
             "yarn",


config.json
```diff
@@ -28,7 +28,7 @@
       }
     },
     "pnpm": {
-      "default": "6.29.1",
+      "default": "7.1.0",
```
arcanis, line 31: Ping @zkochan @aduh95, I'm not sure whether we can/should/must upgrade the default pnpm from 6 to 7, what do you think (I'm worried it would make it a breaking change for Node)?

Also cc @mcollina: with your proposal (automatically retrieving the latest versions), in a situation like this one, do you think Corepack should just upgrade from 6 to 7?

aduh95: I think we should aim for the latest, Corepack is still experimental in Node.js so breaking changes are still expected (and the upside is worth the potential breakage imo).

I think Matteo proposal was we should use a semver-range as the default (so `7.x` in this instance), so on a fresh install of Node.js you're less likely to get a version that contains known security issues.

zkochan: I don't know, whatever the Node.js team thinks is appropriate.
But do update pnpm to v7 in Node.js 18.

ljharb: It should use ^ by default, meaning that updating a major should be considered a breaking change. Exceptions should also be possible, like for npm 8, where the only breaking change wrt 7 was dropping node 10.

aduh95 (approved):
LGTM

config.json
```diff
@@ -1,7 +1,7 @@
 {
   "definitions": {
     "npm": {
-      "default": "8.4.0",
+      "default": "8.10.0",
```
styfle, line 4: Forgive my drive-by comment here, but I thought npm wasn't supported with corepack?

merceyz: It has always been supported just not enabled by default when you run `corepack enable`, you can explicitly enable it with `corepack enable npm`
Ref https://github.com/nodejs/corepack/pull/24

> Note that the npm shims will not be installed unless explicitly requested, as npm is currently distributed with Node.js through other means.
https://github.com/nodejs/corepack/blob/648bd95e947d0a6be0beba4c594de127d94da7d9/README.md#corepack-enable--name

PR #107, opened 2022-05-15
chore: update dependencies

aduh95:
When working on https://github.com/nodejs/corepack/pull/100, I noticed updating `proxy-agent` was blocked by some issue, so I left the outdated version in the other PR that was big enough as is.

Turned out the reason was upgrading `proxy-agent` adds `vm2` to the dependency tree (even though it's never actually used, at least not by the APIs Corepack uses), and it seems that something between `vm2`, Webpack, and PnP is not behaving as expected and the build fails. I used `yarn patch` to basically remove all of `vm2` code to replace it by some minimal mock file, and now Webpack doesn't freak out, and the tests are passing.

I've also bumped the other packages that were left over, and the Yarn version itself.


aduh95:
@arcanis do you know why using a patch version breaks the CI?

arcanis:
Good thing you disabled vm2, it had a vulnerability: https://security.snyk.io/package/npm/vm2

PR #109, opened 2022-05-24
Upgrades to Yarn 4 RC

arcanis:
The Yarn releases are quite stable, even RCs, so I'd like to stay up-to-date when possible.

One major change is that the checked-in cache is now disabled by default since it's only needed in specific situation (typically repositories with large amount of contributions per day); Corepack not being one, I didn't re-enable it in the configuration.


ljharb:
I’m confused; if it’s stable why is it still an RC?

node ships corepack now, and node shouldn’t ship RCs.

arcanis:
> node ships corepack now, and node shouldn’t ship RCs.

I admit the PR title is a little confusing - this PR doesn't change anything to what Corepack distributes (that's what #106 is for). It simply changes the package manager version used to develop on this very repository.

.github/workflows/ci.yml
```diff
@@ -20,16 +20,15 @@ jobs:
         with:
           node-version: 14
 
-      - name: "Check that the cache files are consistent with their remote sources"
-        run: |
-          git diff --exit-code --quiet HEAD^ -- .yarn/cache || \
-            corepack yarn --immutable --immutable-cache --check-cache
-        if: |
-          github.event.pull_request != ''
+      - run: corepack yarn install
```
aduh95, line 23: Non-blocking suggestion: IIUC, those flags are implicitly enable when Yarn 4 is run from a CI env, correct? While I agree it's a nice default, Iimo it's still worth explicitly adding the files so a human can run the command by copy pasting without needing to learn Yarn's defaults.
```suggestion
      - run: corepack yarn install --immutable --refresh-lockfile
```

aduh95 (approved):

.github/workflows/ci.yml
aduh95: Makes sense. We could add a comment explaining that.

.github/workflows/ci.yml
```diff
@@ -53,7 +52,7 @@ jobs:
         with:
           node-version: ${{matrix.node}}.x
 
-      - run: corepack yarn install --immutable
+      - run: corepack yarn install
```
aduh95, line 55: Is this necessary if the previous job has already made the necessary verification?
```suggestion
```

aduh95: Ah gotcha I thought we were still on zero-install.

merceyz (changes requested):
Since the cache isn't committed anymore could you add a `actions/cache` step?

.github/workflows/ci.yml
```diff
@@ -19,17 +19,27 @@ jobs:
       - uses: actions/setup-node@v3
         with:
           node-version: 14
+          
+      - name: Get the Yarn cache directory path
+        id: yarn-cache-dir-path
+        run: echo "::set-output name=dir::$(yarn config get cacheFolder)"
```
merceyz, line 25:
```suggestion
        run: echo "::set-output name=dir::$(corepack yarn config get cacheFolder)"
```

.github/workflows/ci.yml
```diff
@@ -53,7 +63,18 @@ jobs:
         with:
           node-version: ${{matrix.node}}.x
 
+      - name: Get the Yarn cache directory path
+        id: yarn-cache-dir-path
+        run: echo "::set-output name=dir::$(yarn config get cacheFolder)"
```
merceyz, line 68:
```suggestion
        run: echo "::set-output name=dir::$(corepack yarn config get cacheFolder)"
```

PR #115, opened 2022-06-01
chore: bump npm deps

aduh95:
Apply all minor and patch updates available for the semvere-ranges we are targeting.

### DIFF
--- a/yarn.lock
+++ b/yarn.lock
@@ -32,36 +32,36 @@ __metadata:
   linkType: hard
 
 "@babel/core@npm:^7.11.6, @babel/core@npm:^7.12.3, @babel/core@npm:^7.14.3":
-  version: 7.17.10
-  resolution: "@babel/core@npm:7.17.10"
+  version: 7.18.2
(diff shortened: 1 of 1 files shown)


aduh95:
The only reason was I think we want to release a new version soon, and I figure having all deps up-to-date was a good idea. I'm fine with skipping.

aduh95:
Nice, does that mean we are to release a new version today?

arcanis:
I think so, although we don't have a publish workflow yet so it won't be possible to publish to npm 🤔

aduh95:
Micheal set up a `NPM_TOKEN` secret that can let us do that from GitHub Actions. I think we just need to do it, if we can re-use a workflow from somewhere else, that should be pretty straight forward. What were you using before?

aduh95:
Some other nodejs repos are using https://github.com/nodejs/node-core-utils/blob/main/.github/workflows/release-please.yml, which requires the commit messages to follow the [Conventional Commits](https://www.conventionalcommits.org/en/v1.0.0/) convention.

PR #116, opened 2022-06-01
ci: test on all supported Node.js versions

aduh95:
- Use the latest LTS Node.js version available rather than having an arbitrary `14` for `Testing chores` job.
- Add Node.js 16.x and 18.x to the test matrix.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -13,13 +13,11 @@ jobs:
 
     steps:
       - uses: actions/checkout@v3
-        with:
-          fetch-depth: 2
 
       - uses: actions/setup-node@v3
         with:
-          node-version: 14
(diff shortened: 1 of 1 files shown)


arcanis:
My only worry is that the Windows tests tend to timeout regularly - when that happens I restart them, but if there are three different jobs where this can happen it'll become painful 🤔

aduh95:
> My only worry is that the Windows tests tend to timeout regularly - when that happens I restart them, but if there are three different jobs where this can happen it'll become painful 🤔

Note that only a green checks on all three platform on Node.js 14.x is required for a PR to land – but I can see how ignoring red CIs wouldn't be ideal either. Not testing at all on 16.x / 18.x is not ideal either imo.

PR #117, opened 2022-06-01
ci: fix the Sync workflow

aduh95:
Currently the Sync workflow is failing because GitHub Actions do not have permission to commit to the `main` branch. Instead, we can ask it to create a PR, which has the upside of making sure the job is not committing ill-formed data to the `main` branch.

Other changes:

 - Schedule to run the job once a week (I think it's plenty) instead of daily.
 - Use the latest Node.js LTS version rather than v14.x.
 - Use `git show` instead of a temporary file.

### DIFF
--- a/.github/workflows/sync.yml
+++ b/.github/workflows/sync.yml
@@ -3,7 +3,8 @@ name: Version Sync
 on:
   workflow_dispatch:
   schedule:
-    - cron: "0 0 * * *"
+    # Run once a week at 00:05 AM UTC on Sunday.
+    - cron: 5 0 * * 0
 
 jobs:
   build:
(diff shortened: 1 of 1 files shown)


.github/workflows/sync.yml
```diff
@@ -3,7 +3,8 @@ name: Version Sync
 on:
   workflow_dispatch:
   schedule:
-    - cron: "0 0 * * *"
+    # Run once a week at 00:05 AM UTC on Sunday.
```
merceyz, line 6:
```suggestion
    # Run once a week at 00:05 UTC on Sunday.
```
There is no AM or PM with the 24-hour clock

PR #121, opened 2022-06-11
ci: add a publish workflow

arcanis:
This PR adds a workflow to automatically publish new versions each time we modify the `version` field in the `package.json`.

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -1,0 +1,35 @@
+name: Build
+
+on:
+  push:
+    branches: [main]
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+
+    steps:
+      - uses: actions/checkout@v3
+
+      - name: Install Node
+        uses: actions/setup-node@v3
+        with:
+          node-version: lts/*
+
+      - name: Get the Yarn cache directory path
+        id: yarn-cache-dir-path
+        run: echo "::set-output name=dir::$(corepack yarn config get cacheFolder)"
+
+      - uses: actions/cache@v3
+        with:
+          path: ${{steps.yarn-cache-dir-path.outputs.dir}}
+          key: ${{runner.os}}-yarn-${{hashFiles('**/yarn.lock')}}
+          restore-keys: |
+            ${{runner.os}}-yarn-
+
+      - name: Publish the package
+        run:
+          corepack yarn
+          corepack yarn npm publish --tolerate-republish
+        env:
+          YARN_NPM_AUTH_TOKEN: ${{secrets.NPM_TOKEN}}
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -29,7 +29,7 @@ jobs:
 
       - name: Publish the package
         run:
-          corepack yarn
+          corepack yarn install --immutable
           corepack yarn npm publish --tolerate-republish
         env:
           YARN_NPM_AUTH_TOKEN: ${{secrets.NPM_TOKEN}}
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -3,6 +3,7 @@ name: Build
 on:
   push:
     branches: [main]
+    paths: [package.json]
 
 jobs:
   build:
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -1,4 +1,4 @@
-name: Build
+name: Publish releases
 
 on:
   push:
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -41,6 +41,6 @@ jobs:
 
       - run:
           corepack yarn install --immutable
-          corepack yarn npm publish --tolerate-republish
+          corepack yarn npm publish
         env:
           YARN_NPM_AUTH_TOKEN: ${{secrets.NPM_TOKEN}}
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -3,7 +3,6 @@ name: Publish releases
 on:
   push:
     branches: [main]
-    paths: [package.json]
 
 jobs:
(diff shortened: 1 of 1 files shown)


.github/workflows/publish.yml
```diff
@@ -0,0 +1,35 @@
+name: Build
+
+on:
+  push:
+    branches: [main]
```
aduh95, line 5:
```suggestion
    branches: [main]
    paths: [package.json]
```

.github/workflows/publish.yml
```diff
@@ -0,0 +1,35 @@
...
+      - uses: actions/cache@v3
+        with:
+          path: ${{steps.yarn-cache-dir-path.outputs.dir}}
+          key: ${{runner.os}}-yarn-${{hashFiles('**/yarn.lock')}}
+          restore-keys: |
+            ${{runner.os}}-yarn-
+
+      - name: Publish the package
+        run:
+          corepack yarn install --immutable
```
aduh95, line 32: We want to publish the built files, right?
```suggestion
          corepack yarn install --immutable
          corepack yarn build
```

arcanis: It's automatically called as part of the [`prepack` script](https://github.com/nodejs/corepack/blob/main/package.json#L60)

.github/workflows/publish.yml
```diff
@@ -0,0 +1,36 @@
...
+          corepack yarn npm publish --tolerate-republish
+        env:
+          YARN_NPM_AUTH_TOKEN: ${{secrets.NPM_TOKEN}}
```
aduh95, line 36: Let's create a GitHub release as well. This will also automatically create a git tag for the release.
```suggestion
          YARN_NPM_AUTH_TOKEN: ${{secrets.NPM_TOKEN}}
      - name: Get version number
        id: versionNumber
        run: jq -r '"##[set-output name=version;]"+.version' < package.json
      - name: Create GitHub release
        run: gh release create v${{ steps.versionNumber.outputs.version }} -t "Corepack ${{ steps.versionNumber.outputs.version }}"  --prerelease
        env:
          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
```

aduh95: Hum but we don't want to create a release if it turns out the commit is not a release actually. Let me check what happens if one tries to release a version that's already been released.

arcanis: The `--tolerate-republish` makes Yarn not do anything, although if we also create a GH release then it won't be enough

aduh95: If we use release-please, that's being taken care of for us, we don't have to bother.

.github/workflows/publish.yml
```diff
@@ -0,0 +1,46 @@
...
+    branches: [main]
+    paths: [package.json]
+
+jobs:
+  release-please:
+    runs-on: ubuntu-latest
+    outputs:
+      release_created:  ${{ steps.release.outputs.release_created }}
+    steps:
+      - uses: google-github-actions/release-please-action@v3
```
aduh95, line 14: Are we OK with using a `v3` instead of a hash?

.github/workflows/publish.yml
```diff
@@ -0,0 +1,46 @@
...
+          path: ${{steps.yarn-cache-dir-path.outputs.dir}}
+          key: ${{runner.os}}-yarn-${{hashFiles('**/yarn.lock')}}
+          restore-keys: |
+            ${{runner.os}}-yarn-
+
+      - run:
+          corepack yarn install --immutable
+          corepack yarn npm publish
+        env:
+          YARN_NPM_AUTH_TOKEN: ${{secrets.NPM_TOKEN}}
```
aduh95, line 46: Side note: it'd be nice to have this env variable documented in https://yarnpkg.com/cli/npm/publish, I didn't know about it and was using `corepack yarn config set npmAuthToken ${{ toJSON(secrets.NPM_TOKEN) }}` on my own repos.

.github/workflows/publish.yml
arcanis: Given in this case it's a Google-owned repository the risk is probably acceptable, compared to an arbitrary GH user

.github/workflows/publish.yml
arcanis: Almost all Yarn settings can be set via environment variables (we document it at the top of the [settings page](https://yarnpkg.com/configuration/yarnrc), but perhaps an explicit callout in the publish command would be worthwhile indeed 🤔)

.github/workflows/publish.yml
aduh95: Hum I'm not sure trusting Google more than individuals is very fair, I think it makes more sense to draw the line at GitHub (because they're the ones who manage the runner anyway) and treat any other non-Node.js with more caution. No strong feelings either way though.

.github/workflows/publish.yml
aduh95: Ah yeah, I completely missed the paragraph about env variables. Maybe there could be a third tab named "Environment variable" which would say something like "Any settings can be set from env variable using this pattern, checkout `.yarnrc` tab for the full list of supported settings."

aduh95 (approved):

.github/workflows/publish.yml
```diff
@@ -0,0 +1,46 @@
+name: Build
```
aduh95, line 1:
```suggestion
name: Publish releases
```

.github/workflows/publish.yml
```diff
@@ -0,0 +1,46 @@
+name: Publish releases
+
+on:
+  push:
+    branches: [main]
+    paths: [package.json]
```
aduh95, line 6: We need to remove that now that we're using release-please.
```suggestion
```

aduh95 (approved):
Looking good!

PR #124, opened 2022-06-12
fix(ci): YAML formatting in publish workflow

aduh95:


arcanis (approved):

aduh95:
@arcanis Can you investigate the failure on the release script? I have to log off now, it'd be nice if we had a version on npm that could then be pulled in time for Node.js 18.4.0.

PR #125, opened 2022-06-12
Fixes publish script

arcanis:
The publish script uses `run:` instead of `run: |`, which causes Yaml to discard line returns


arcanis:
Duplicate of #124

Issue #37, opened 2021-06-13
Allow checking hashes for exact versions, and check them for default versions


andersk:
Corepack currently depends on the security of HTTPS. Projects that want stronger guarantees should optionally be allowed to configure an integrity hash for the downloaded package manager. Obviously this would only be possible when configuring an exact version rather than a range.

Integrity hashes for the default known-good versions should be included in Corepack, so that users get the strongest security guarantees by default.

(Related to nodejs/corepack#10, but without requiring modifications to the registry or the package managers.)

PR #133, opened 2022-06-24
feat: add support for hash checking

aduh95:
```console
$ corepack yarn@3.2.1+sha256.deadbeef
Internal Error: Mismatch hashes. Expected deadbeef, got 4e66b50e036d75c28ccd3fb96350bf3901b0a88e4f561da6c2ad5109552a7380
    at IncomingMessage.<anonymous> (/Users/duhamean/Documents/corepack/dist/corepack.js:15884:24)
    at IncomingMessage.emit (node:events:549:35)
    at endReadableNT (node:internal/streams/readable:1359:12)
    at process.processTicksAndRejections (node:internal/process/task_queues:82:21)
```

I'm not sure it's how we want to do it, according to semver.org, we could use build suffix to the version number (e.g. `yarn@3.2.1+hash`), ~that would force us to pick a hashing algorithm (which is not a problem in itself, but could create issues in the future if it becomes outdated)~.

EDIT: with a new version we're now using semver compatible suffix for the hash.

Fixes: https://github.com/nodejs/corepack/issues/37


aduh95:
> > that would force us to pick a hashing algorithm
> 
> Not necessarily, you can include the algorithm in the build metadata as well.
> 
> ```
> yarn@3.2.1+sha512.hash
> ```

Ah yes, that’s right – IIUC dot is not a valid character, but we could use a dash (unless there are algorithms using a dash in their name?)

merceyz:
A dot is perfectly valid, see the https://regex101.com/r/vkijKf/1/ (from https://semver.org/#is-there-a-suggested-regular-expression-regex-to-check-a-semver-string)

sources/httpUtils.ts
```diff
@@ -27,28 +42,41 @@ export async function fetchUrlStream(url: string, options: RequestOptions = {})
   });
 }
 
-export async function fetchAsBuffer(url: string, options?: RequestOptions) {
+export async function fetchAsBuffer(url: string, hash?: string, options?: RequestOptions) {
```
arcanis, line 45: nit: I'd put the hash as part of the option bag

arcanis (approved):
Quite elegant solution 👍

sources/httpUtils.ts
aduh95: I didn't do it because `options` is passed to the Node.js `https.get` method, and I thought it could be problematic if it clashes in the future with one of Node.js' supported options – and if I'm being honest, I was feeling too lazy to update the types.

aduh95:
I just realised I'm checking the hash of the JSON metadata instead of the tarball 🤦‍♂️ Converting to draft for now.

sources/httpUtils.ts
aduh95: I've refactored that away, I now rely on semver to parse the build info and removed the additional argument.

sources/Engine.ts
```diff
@@ -127,7 +127,7 @@ export class Engine {
       throw new UsageError(`This package manager (${descriptor.name}) isn't supported by this corepack build`);
 
     let finalDescriptor = descriptor;
-    if (descriptor.range.match(/^[a-z-]+$/)) {
+    if (/^[a-z-]+?$/.test(descriptor.range)) {
```
aduh95, line 130:
```suggestion
    if (/^[a-z-]+$/.test(descriptor.range)) {
```

sources/Engine.ts
```diff
@@ -151,6 +151,11 @@ export class Engine {
     if (cachedVersion !== null && useCache)
       return {name: finalDescriptor.name, reference: cachedVersion};
 
+    // If the user asked for a specific version, no need to request the list of
+    // available versions from the registry.
+    if (semver.satisfies(finalDescriptor.range, finalDescriptor.range))
```
aduh95, line 156:
```suggestion
    if (semver.valid(finalDescriptor.range))
```

tests/main.test.ts
```diff
@@ -8,13 +8,38 @@ beforeEach(async () => {
...
+    await expect(runCli(cwd, [`yarn`, `--version`])).resolves.toMatchObject({
+      exitCode: 1,
+      stdout: /Mismatch hashes/,
+    });
+  });
+});
+
 const testedPackageManagers: Array<[string, string]> = [
   [`yarn`, `1.22.4`],
+  [`yarn`, `1.22.4+md5.faf483d50aa8ccbdc802efa0cac5d4d3`],
```
arcanis, line 26: nit: we don't test `sha1`, which is what #137 uses

arcanis (approved):

tests/main.test.ts
aduh95: Makes sense, I've switched to SHA1 and SHA224 in the tests as those are more credible options for hash algorithms.

Issue #93, opened 2022-02-17
Dynamic default versions


mcollina:
One of the main reason we adopted corepack was to avoid work in nodejs/node to support multiple package managers. There is a measurable amount of traffic on our H1 due to vulnerabilities in the npm dependencies, and we are struggling to keep npm up to date. Unfortunately corepack increases this load and it makes it even less apparent and hard to track as the vulnerabilities would not be apparent to the users.

I propose that the config file include only the major version of each package manager. This will ensure that for a specific version of Node.js there won't be any breaking change (I consider the drop of a Node.js version a breaking change).
Then we will update the major version of each package manager whenever we ship a new major release of Node.js.

I would also recommend that the config file is not bundled in but loaded at runtime. This will simplify maintenance on the Node.js side so we could just update the config file without updating corepack (and viceversa). This also enables mulitple Node.js lines to have the same version of corepack but different defaults.

arcanis:
It could be interesting to make this a GH discussion, since there are multiple parallel discussion points.

> There is a measurable amount of traffic on our H1 due to vulnerabilities in the npm dependencies

Out of curiosity, do you have stats on how many of those vulnerabilities were truly relevant, vs "rubber stamp upgrade" intended to silence a noisy report?

> We are struggling to keep npm up to date

I'm really new to the Node infra and procedures so feel free to correct me, but isn't it something that should be "easy" to automate? What exactly makes this process difficult (I'd like to better understand that to take it into account in future design iterations)?

> I propose that the config file include only the major version of each package manager. This will ensure that for a specific version of Node.js there won't be any breaking change.

This isn't guaranteed. Package managers can make mistakes too, and each upgrades has the potential to ship a bug that would break someone's workflow (as an example, I remember npm accidentally shipped a few years ago a rc which erased the home folder in certain circumstances). Even bug fixes [can be breaking changes](https://xkcd.com/1172/), depending on the perspective.

---

In the case of Yarn we don't really recommend people to use the global version anyway, so technically speaking it wouldn't affect us much; my concern more around how Node distributors would see the perspective of shipping non-deterministic code to their users (even if the package managers are currently lazily installed, they are still deterministic).

For example, should Corepack be enabled by default, do you believe you'd be able to convince Ubuntu to ship the Corepack shims with the logic you describe? Or would it effectively prevent Corepack from ever being enabled by default?

arcanis:
[...]

Issue #10, opened 2020-09-29
Signing of distributions


jasnell:
Ideally, the utility would include the requirement that package manager distributions are signed, with the ability for users to verify the signature on installation.

arcanis:
That would be neat, especially since each time could bake their public key in Corepack 🤔

However, while we could sign the Yarn 2 releases by writing an header comment containing the signature, it wouldn't be possible for package managers installed from the npm registry tarballs.

The main way I could see it work would be to have some kind of deterministic archive traversal that would compute the signature for all files from the archive minus the signature file, and would compare it to the signature file stored within the archive (that would be added right after packing). However,

- it would require pnpm and npm (optional, since we don't officially expose their entrypoint) to build the relevant infra
- changes in the way the archives are traversed could lead to signature mismatches - I'm worried that would be prone to fail

Apart from that, the only option would be for the npm registry to support detached signatures.

andersk:
Verifying a SHA-256 hash (#37) would provide much of this benefit without any modifications to the way package managers are distributed.

aduh95:
AKAICT, none of the package managers that Corepack currently supports sign their releases. If Yarn is the easiest one to sign, could we start with this one? @arcanis do you know if that would be difficult to add a step for that in the Yarn release workflow?

aduh95:
Sigstore seems to be npm package specific, I'm not sure if it would work well for us 🤔

I was thinking on what would be a generic solution, and maybe the following would work: the package manager author would bundle a [policy manifest](https://nodejs.org/api/permissions.html) as well as an RSA signature for said manifest file. The upside of that approach is that it would allow the package manager to fine grain the security of their application in a way that's simply not possible currently. wdyt?

RichiCoder1:
@aduh95 ah! To be clear, that library is calling out and supporting npm packages specifically because it's what the new [npm provenance](https://github.com/github/roadmap/issues/657) feature is built on.

It will work with any file though, and is currently used for verification in other popular formats like [containers](https://docs.sigstore.dev/cosign/overview/) and has generic documentation for [other formats](https://docs.sigstore.dev/cosign/overview/#other-formats).

ljharb:
Note that that provenance feature adds nothing that npm hasn't had for a decade - although using sigstore might make it more agnostic.

MylesBorins:
> Note that that provenance feature adds nothing that npm hasn't had for a decade - although using sigstore might make it more agnostic.

[...]

PR #134, opened 2022-06-26
feat: download the latest version instead of a pinned one

aduh95:
This PR ensures that fresh install of Corepack would always use the latest available version for a given package manager.

We probably want to add a command to clear the cache for `lastKnownGood` version (although I think that's more or less what `corepack prepare` is doing), and/or have a expiracy date for the `lastKnownGood` version to make Corepack update the default version from time to time.

Fixes: https://github.com/nodejs/corepack/issues/93

### DIFF
--- a/README.md
+++ b/README.md
@@ -46,9 +46,16 @@ Just use your package managers as you usually would. Run `yarn install` in Yarn 
 
 ## Known Good Releases
 
-When running Yarn or pnpm within projects that don't list a supported package manager, Corepack will default to a set of Known Good Releases. In a way, you can compare this to Node.js, where each version ships with a specific version of npm.
+When running Corepack within projects that don't list a supported package
+manager, it will default to a set of Known Good Releases. In a way, you can
+compare this to Node.js, where each version ships with a specific version of npm.
 
-The Known Good Releases can be updated system-wide using the `--activate` flag from the `corepack prepare` and `corepack hydrate` commands.
+If there is no Known Good Release for the requested package manager, Corepack
+looks up the npm registry for the latest available version and cache it for
+future use.
+
+The Known Good Releases can be updated system-wide using the `--activate` flag
+from the `corepack prepare` and `corepack hydrate` commands.
 
 ## Offline Workflow
 
--- a/README.md
+++ b/README.md
@@ -106,6 +106,8 @@ This command will retrieve the given package manager from the specified archive 
 
 - `COREPACK_ENABLE_NETWORK` can be set to `0` to prevent Corepack from accessing the network (in which case you'll be responsible for hydrating the package manager versions that will be required for the projects you'll run, using `corepack hydrate`).
 
+- `COREPACK_NO_LOOKUP` can be set in order to instruct Corepack not to lookup on the remote registry for the latest version of the selected package manager.
+
 - `COREPACK_HOME` can be set in order to define where Corepack should install the package managers. By default it is set to `$HOME/.node/corepack`.
 
 - `COREPACK_ROOT` has no functional impact on Corepack itself; it's automatically being set in your environment by Corepack when it shells out to the underlying package managers, so that they can feature-detect its presence (useful for commands like `yarn init`).
--- a/README.md
+++ b/README.md
@@ -113,7 +113,9 @@ This command will retrieve the given package manager from the specified archive 
 
 - `COREPACK_ENABLE_NETWORK` can be set to `0` to prevent Corepack from accessing the network (in which case you'll be responsible for hydrating the package manager versions that will be required for the projects you'll run, using `corepack hydrate`).
 
-- `COREPACK_NO_LOOKUP` can be set in order to instruct Corepack not to lookup on the remote registry for the latest version of the selected package manager.
+- `COREPACK_DEFAULT_TO_LATEST` can be set to `0` in order to instruct Corepack
+  not to lookup on the remote registry for the latest version of the selected
+  package manager.
 
 - `COREPACK_HOME` can be set in order to define where Corepack should install the package managers. By default it is set to `$HOME/.node/corepack`.
 
--- a/sources/Engine.ts
+++ b/sources/Engine.ts
@@ -1,6 +1,7 @@
 import {UsageError}                                           from 'clipanion';
 import fs                                                     from 'fs';
 import path                                                   from 'path';
+import process                                                from 'process';
 import semver                                                 from 'semver';
 
 import defaultConfig                                          from '../config.json';
@@ -77,7 +78,7 @@ export class Engine {
       }
     }
 
-    if (process.env.COREPACK_NO_LOOKUP)
+    if (process.env.COREPACK_DEFAULT_TO_LATEST === `0`)
       return definition.default;
 
     const reference = await corepackUtils.fetchLatestStableVersion(definition.fetchLatestFrom);
--- a/sources/Engine.ts
+++ b/sources/Engine.ts
@@ -7,6 +7,7 @@ import defaultConfig                                          from '../config.js
 
 import * as corepackUtils                                     from './corepackUtils';
 import * as folderUtils                                       from './folderUtils';
+import {fetchAsJson}                                          from './httpUtils';
 import * as semverUtils                                       from './semverUtils';
 import {Config, Descriptor, Locator}                          from './types';
 import {SupportedPackageManagers, SupportedPackageManagerSet} from './types';
@@ -69,17 +70,16 @@ export class Engine {
       // Ignore errors; too bad
     }
 
-    if (typeof lastKnownGood !== `object` || lastKnownGood === null)
-      return definition.default;
+    if (typeof lastKnownGood === `object` && lastKnownGood !== null &&
+        Object.prototype.hasOwnProperty.call(lastKnownGood, packageManager)) {
+      const override = (lastKnownGood as any)[packageManager];
+      if (typeof override === `string`) {
+        return override;
+      }
+    }
 
-    if (!Object.prototype.hasOwnProperty.call(lastKnownGood, packageManager))
-      return definition.default;
-
-    const override = (lastKnownGood as any)[packageManager];
-    if (typeof override !== `string`)
-      return definition.default;
-
-    return override;
+    const latest = await fetchAsJson(`https://registry.npmjs.org/${packageManager}`);
+    return latest[`dist-tags`].latest;
   }
 
   async activatePackageManager(locator: Locator) {
--- a/sources/Engine.ts
+++ b/sources/Engine.ts
@@ -81,8 +81,15 @@ export class Engine {
     if (process.env.COREPACK_NO_LOOKUP)
       return definition.default;
 
-    const latest = await fetchAsJson(`https://registry.npmjs.org/${packageManager}`);
-    return latest[`dist-tags`].latest;
+    const {[`dist-tags`]: {latest}, versions: {[latest]: {dist: {shasum}}}} = await fetchAsJson(`https://registry.npmjs.org/${packageManager}`);
+    const reference = `${latest}+sha1.${shasum}`;
+
+    await this.activatePackageManager({
+      name: packageManager,
+      reference,
+    });
+
+    return reference;
   }
 
   async activatePackageManager(locator: Locator) {
--- a/sources/Engine.ts
+++ b/sources/Engine.ts
@@ -78,6 +78,9 @@ export class Engine {
       }
     }
 
+    if (process.env.COREPACK_NO_LOOKUP)
+      return definition.default;
+
     const latest = await fetchAsJson(`https://registry.npmjs.org/${packageManager}`);
     return latest[`dist-tags`].latest;
   }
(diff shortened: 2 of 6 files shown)


sources/Engine.ts
```diff
@@ -69,17 +70,19 @@ export class Engine {
...
+    }
 
-    const override = (lastKnownGood as any)[packageManager];
-    if (typeof override !== `string`)
+    if (process.env.COREPACK_NO_LOOKUP)
       return definition.default;
 
-    return override;
+    const latest = await fetchAsJson(`https://registry.npmjs.org/${packageManager}`);
+    return latest[`dist-tags`].latest;
```
arcanis, line 85: This fetch would be executed at every invocation, it'd be quite slow for things like `yarn run some-script` which themselves call other scripts.

I think it'd need to be cached at least on a by-process basis (through an env variable that would be set when spawning the subprocess, for example `COREPACK_LKG_YARN=1.22.18`), and possibly in a system-wide cache (for example to only fetch the new versions every 24 hours).

aduh95: Wouldn't the last downloaded version be cached by `lastKnownGood`?

sources/Engine.ts
```diff
@@ -69,17 +70,26 @@ export class Engine {
...
+    }
 
-    if (!Object.prototype.hasOwnProperty.call(lastKnownGood, packageManager))
+    if (process.env.COREPACK_NO_LOOKUP)
       return definition.default;
 
-    const override = (lastKnownGood as any)[packageManager];
-    if (typeof override !== `string`)
-      return definition.default;
+    const {[`dist-tags`]: {latest}, versions: {[latest]: {dist: {shasum}}}} = await fetchAsJson(`https://registry.npmjs.org/${packageManager}`);
```
merceyz, line 84: We can't hardcode the npm registry here, it needs to look at what `config.json` specifies.
See the following section for an example https://github.com/nodejs/corepack/blob/31913aa3172e8c3fe65528996f9b8f32a43b52a6/sources/corepackUtils.ts#L12-L26

README.md
```diff
@@ -46,9 +46,16 @@ Just use your package managers as you usually would. Run `yarn install` in Yarn
 
 ## Known Good Releases
 
-When running Yarn or pnpm within projects that don't list a supported package manager, Corepack will default to a set of Known Good Releases. In a way, you can compare this to Node.js, where each version ships with a specific version of npm.
+When running Corepack within projects that don't list a supported package
+manager, it will default to a set of Known Good Releases. In a way, you can
+compare this to Node.js, where each version ships with a specific version of npm.
 
-The Known Good Releases can be updated system-wide using the `--activate` flag from the `corepack prepare` and `corepack hydrate` commands.
+If there is no Known Good Release for the requested package manager, Corepack
```
styfle, line 53: In what case would there be no Known Good Release?

README.md
```diff
@@ -106,6 +113,8 @@ This command will retrieve the given package manager from the specified archive
 
 - `COREPACK_ENABLE_NETWORK` can be set to `0` to prevent Corepack from accessing the network (in which case you'll be responsible for hydrating the package manager versions that will be required for the projects you'll run, using `corepack hydrate`).
 
+- `COREPACK_NO_LOOKUP` can be set in order to instruct Corepack not to lookup on the remote registry for the latest version of the selected package manager.
```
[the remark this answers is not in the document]
styfle: I'm trying to understand the difference between `COREPACK_ENABLE_NETWORK=0` and `COREPACK_NO_LOOKUP=1`? In theory I would expect no network to behave the same since corepack ships with Known Good Versions but I must be overlooking a use case here.

README.md
aduh95: Upon the first time Corepack is run with a given package manager. The idea behind this PR would be not to rely on what version is set on `config.json` but instead get the most up-to-date version from npm registry.

README.md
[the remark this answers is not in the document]
aduh95: When `COREPACK_NO_LOOKUP` is set, Corepack will use the version defined in `config.json` instead of asking the npm registry for the most up-to-date version. `COREPACK_ENABLE_NETWORK` can disable all access to the network, both for finding what's the most up-to-date version number and to download the tarball.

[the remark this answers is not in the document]
styfle: Is `config.json` available to configure by the user of corepack or is it just internals?

README.md
styfle: How often does it check the registry? Does it check every time you invoke yarn/pnpm/npm?

README.md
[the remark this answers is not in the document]
aduh95: It's only internals, bundled in the corepack executable.

README.md
aduh95: No, only the first time, at which point it saves it as Known Good Release. Users can then update (or downgrade) that version by using `corepack prepare` or `corepack hydrate` (e.g. `corepack prepare pnpm@latest --activate` to get the latest pnpm by default). We could make an automated process for the Known Good Version to expire after some time, but that's not part of this initial implementation.

styfle: Thanks, I understand now 👍

I think expiring would could be confusing. I also think this any auto update could be confusing because it no longer ties the package manager version to the corepack version (or node version for that matter). 

I think it’s best to let the user decide when to upgrade and perhaps provide better docs how to upgrade (this isn’t super clear today). Without corepack, this is handled with a notification that says something like npm i -g npm@latest. 

Has any of this behavior been reviewed by the npm team? I think they had some ideas about Know Good Release.

aduh95: > I also think this any auto update could be confusing because it no longer ties the package manager version to the corepack version (or node version for that matter).

Well that's the reason Corepack exists, because the Node.js TSC doesn't want to have package managers tied in with a specific Node.js version – specifically, several Node.js TSC members have expressed that the reason they are suportive on Corepack is to get rid of having to do security releases every time a vulnerability is reported in npm rather than `node` itself.
So the auto-lookup during the first use _has to be_ the default behavior.


> I think it’s best to let the user decide when to upgrade and perhaps provide better docs how to upgrade (this isn’t super clear today). Without corepack, this is handled with a notification that says something like npm i -g npm@latest.

Yes that makes sense, letting the end-user in control is very much the goal here. So instead of `npm i -g npm@latest`, users would type `corepack prepare npm@latest --activate`, which is arguably quite similar. Improving docs is of course always a goal.

> Has any of this behavior been reviewed by the npm team? I think they had some ideas about Know Good Release.

The npm team hasn't shown much interest in the development of Corepack, so I don't think they have.

README.md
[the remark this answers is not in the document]
styfle: How about COREPACK_IGNORE_LATEST?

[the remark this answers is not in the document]
aduh95: I went with `COREPACK_DEFAULT_TO_LATEST` which is true by default.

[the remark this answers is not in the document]
styfle: Not sure about using default as a verb here. 

How about COREPACK_ENABLE_FETCH_LATEST=0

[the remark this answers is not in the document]
aduh95: I'd like the naming to make clear that it's only relevant if there's no local clue on what version to use. `ENABLE_FETCH_LATEST` could mean that the latest is always downloaded, or at least that's how I read it.

> Not sure about using “default” as a verb here.

I'm not a native english speaker, so genuine question: is it like incorrect to use "default" as a verb?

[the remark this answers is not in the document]
styfle: It’s not incorrect, just that “default” can be a verb or a noun. It’s probably fine 👍

sources/corepackUtils.ts
```diff
@@ -12,6 +12,23 @@ import * as httpUtils                                          from './httpUtils
 import * as nodeUtils                                          from './nodeUtils';
 import {RegistrySpec, Descriptor, Locator, PackageManagerSpec} from './types';
 
+export async function fetchLatestStableVersion(spec: RegistrySpec) {
+  switch (spec.type) {
+    case `npm`: {
+      const {[`dist-tags`]: {latest}, versions: {[latest]: {dist: {shasum}}}} =
```
styfle, line 18: Should this consider the major version? What about when npm 7 bumped to npm 8 to drop support for node 12. Imagine if corepack existed then and suddenly npm stopped worked because of the automatic fetch latest.

aduh95: That’s worth discussing, but imo it’s ok to ignore that for the initial implementation.

arcanis (approved):
I think that's acceptable to me (Yarn isn't impacted *that much* since we don't make very frequent / risky releases to 1.22 anyway). Maybe we should wait a bit to see if @zkochan has an opinion?

zkochan:
Sorry for not answering. I think it is OK to ship the latest pnpm. Thanks

andersk:
I have some concerns about this change.

1. This negates the hash protection of #137.

2. This could break the reproducibility of builds or CI jobs that start from a specific Node.js version (unless specifically overridden).

3. Consider a user who gets semi-regular automatic Node.js updates from their distribution, which means they get semi-regular automatic corepack updates. Under the previous scheme, they would therefore get semi-regular automatic updates to the default versions of Yarn and pnpm. But now those versions will be fixed from the first time they’re run (until a manual update).

aduh95:
> 1. This negates the hash protection of [chore: add sha1 to default versions when available #137](https://github.com/nodejs/corepack/pull/137).
> 2. This could break the reproducibility of builds or CI jobs that start from a specific Node.js version (unless specifically overridden).

The workaround is to set a specific version in the project `package.json` and/or set `COREPACK_DEFAULT_TO_LATEST` to `0` in the CI env.


> 3\. Consider a user who gets semi-regular automatic Node.js updates from their distribution, which means they get semi-regular automatic corepack updates. Under the previous scheme, they would therefore get semi-regular automatic updates to the default versions of Yarn and pnpm. But now those versions will be fixed from the first time they’re run (until a manual update).

Contributions are welcome to improve this workflow.

andersk:
Of course there are workarounds. But good defaults matter, and I’m proposing that the default should remain the secure known-good version.

A user who discovers they need a later version than the known-good version is in a position to go read the corepack documentation and configure it appropriately. (Most likely what they really want is to set a version in `package.json` so that everyone working on that project gets the right version. But if they want to set it system-wide, sure, they can have that option.)

Meanwhile, a user who unexpectedly gets a later (or earlier!) version when they *didn’t make a change*—they just re-ran the same script on a different system—is in a much more confusing position. They could very well be trying to get dozens of things working at the same time; they might not be an expert on the particular part of the script that failed, and might not even realize in the moment that their Yarn or pnpm is coming from corepack. Could they (or someone else who wrote the script) have had the foresight to reconfigure corepack more deterministically? Of course, but a trap that could be avoided is still a trap; good defaults matter.

Users don’t expect that installing a deterministic version of A should give them a nondeterministic version of B. We shouldn’t force them to remember or disable an exception to this rule.

ljharb:
That’s exactly what users in this ecosystem expect - determinism, for better or worse, is simply not an expectation (beyond a lockfile), and anything else is surprising and very inconvenient, and in this case, causes many more people to have a buggy or insecure version of their package manager CLI.

aduh95:
> But good defaults matter, and I’m proposing that the default should remain the secure known-good version.

That's the catch, the known-good version is more likely to be less secure than the latest one.

I agree with Jordan, I don't think most users care about that kind of determinism (otherwise, they would set a specific version either system-wise, or at the project level). Also, as I wrote in https://github.com/nodejs/corepack/pull/134?notification_referrer_id=NT_kwDOANpZjbMzODgzNzkyMzM1OjE0MzA5Nzcz#discussion_r917276616, the reason Corepack is integrated in the Node.js project is that it make decouple what version of npm comes bundled with the Node.js installation, users can get a secure version of their package manager without Node.js needing to issue a security update.

styfle:
> I don't think most users care about that kind of determinism (otherwise, they would set a specific version either system-wise

What makes you think that users are not setting a specific version? I see many projects using `npm i -g pnpm@7.9.5`. In fact, package managers print a warning and tell the user the upgrade command when running an outdated version.

andersk:
This change isn’t giving them the latest version. It’s giving them the latest version *as of* the time they ran `corepack enable`. That could be ages earlier than the time they upgraded Node.js and corepack, since it’s a different tool they could easily forget to run regularly when upgrading everything else.

But the security concern isn’t just that they might get an older version—it’s that skipping the hash protection provided by known-good versions may make it easier for an attacker to maliciously alter the code in transit, since HTTPS can much easier to defeat in environments with intercepting proxy middleboxes and dubious certificate authorities.

aduh95:
> This change isn’t giving them the latest version. It’s giving them the latest version _as of_ the time they ran `corepack prepare`. That could be ages earlier than the time they upgraded Node.js and corepack, since it’s a different tool they could easily forget to run regularly when upgrading everything else.

I reckon the current setup is not ideal, it's a first implementation and I'd be delighted to see it improve in a way that's convenient for the users. Contributions would be much welcome, as we don't have a clear plan on how to give a way for users to update without relying on them remembering to run a command manually on a regular basis.



> But the security concern isn’t just that they might get an older version—it’s that skipping the hash protection provided by known-good versions may make it easier for an attacker to maliciously alter the code in transit, since HTTPS can much easier to defeat in environments with intercepting proxy middleboxes and dubious certificate authorities.

I think https://github.com/nodejs/corepack/issues/10 would be a better response to this, as it should give us the best of both worlds.

PR #135, opened 2022-06-29
fix: streamline cache exploration

aduh95:
If a large number of versions are cached on the host system, trying to read the whole cache at once can be quite memory intensive. Instead, we can use an iterative approach that doesn't hold the whole directory content in memory.


arcanis (approved):
Looks ok - do you have before/after numbers, for reference?

aduh95:
> Looks ok - do you have before/after numbers, for reference?

I don't have any, and tbh I would expect for it to only matter if the cache directory grows significantly on a memory limited system. But better safe than sorry, right?

Issue #145, opened 2022-07-13
Follow XDG Specification


shubham-cpp:
[XDG Base Directory](https://wiki.archlinux.org/title/XDG_Base_Directory)

By default, corepack creates folder on `$HOME/.node/corepack`. This should not be the default, but rather a fallback.
Default location should be something like `$XDG_CACHE_HOME/node/corepack` or `$HOME/.cache/node/corepack` respecting the xdg specifications

ljharb:
Note that the XDG specification uses "should" liberally, and "must" only rarely, so there's very little the XDG specification actually requires things to do in order to be compliant.

shubham-cpp:
Agreed, but since npm, pnpm and yarn,nvm,ect also provide options for xdg specs, then why not provide an option for corepack as well. People who might want to opt in to xdg specs can use that(like environment variable)

aduh95:
Is it something you'd like to open a PR for?

shubham-cpp:
Need to look into this

Mitsunee:
came here to open this exact issue. I have XDG user dirs configured and

```
prefix=${XDG_DATA_HOME}/node
cache=${XDG_CACHE_HOME}/npm
init-module=${XDG_CONFIG_HOME}/npm/config/npm-init.js
```

in my npmrc (location configured via `$NPM_CONFIG_USERCONFIG`, which is fine) and now corepack and cacache are the two things ignoring this (yarn at least mostly follows XDG dirs) and constantly recreating the `~/.npm` and `~/.node` directories I'm trying to avoid having to a completely clean setup. I opened a ticket for cacache already here https://github.com/npm/cacache/issues/119 and it'd be great if corepack could also use the `$XDG_DATA_HOME` variable if available.

shubham-cpp:
Big thank you to all the guys that worked on this.
Open source community at its best again.

PR #152, opened 2022-07-23
feat: do not use `~/.node` as default value for `COREPACK_HOME`

aduh95:
Fixes: https://github.com/nodejs/corepack/issues/145
Refs: https://github.com/nodejs/node/issues/43859

### DIFF
--- a/sources/folderUtils.ts
+++ b/sources/folderUtils.ts
@@ -10,8 +10,8 @@ export function getInstallFolder() {
     process.env.COREPACK_HOME ??
     join(
       process.env.XDG_CACHE_HOME ??
-        process.env.APPDATA ??
-        join(homedir(), `.cache`),
+        process.env.LOCALAPPDATA ??
+        join(homedir(), process.platform === `win32` ? `AppData/Local` : `.cache`),
       `node/corepack`,
     )
   );
--- a/sources/folderUtils.ts
+++ b/sources/folderUtils.ts
@@ -1,11 +1,32 @@
-import {mkdirSync}       from 'fs';
-import {homedir, tmpdir} from 'os';
-import {join}            from 'path';
-import process           from 'process';
+import {existsSync, mkdirSync, renameSync} from 'fs';
+import {homedir, tmpdir}                   from 'os';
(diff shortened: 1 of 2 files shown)


sources/folderUtils.ts
```diff
@@ -1,10 +1,20 @@
...
+import type {NodeError}  from './nodeUtils';
 
 export function getInstallFolder() {
-  return process.env.COREPACK_HOME ?? join(homedir(), `.node/corepack`);
+  return (
+    process.env.COREPACK_HOME ??
+    join(
+      process.env.XDG_CACHE_HOME ??
+        process.env.APPDATA ??
+        join(homedir(), `.cache`),
```
merceyz, line 14:
```suggestion
        process.env.LOCALAPPDATA ??
        join(homedir(), process.platform === `win32` ? `AppData/Local` : `.cache`),
```
Local seems more appropriate here.

merceyz (changes requested):
This will essentially "clear" the state of Corepack on existing machines, we should probably see if `~/.node/corepack` already exists and continue using it or "migrate" its data to the new location.

sources/folderUtils.ts
```diff
@@ -1,10 +1,41 @@
...
+  }
+  return (
+    process.env.COREPACK_HOME ??
+    join(
+      process.env.XDG_CACHE_HOME ??
+        process.env.LOCALAPPDATA ??
+        join(homedir(), process.platform === `win32` ? `AppData/Local` : `.cache`),
+      `node/corepack`,
+    )
+  );
```
arcanis, line 38: You should swap the logic so that this part is the early return, since it's the shortest - also I think the `??` here is useless, because you already checked for `process.env.COREPACK_HOME == null` earlier 🤔

aduh95: The idea is to remove the block above in the next major (see TODO line 10). It's written like this so it's easier to remove the "migration path" that we definitely don't want to keep around for too long.

arcanis: Fair enough 👍

arcanis (approved):

Issue #157, opened 2022-08-07
Usage Error: This project is configured to use <pkgmgr>


styfle:
One of the biggest problems I have today with Corepack is the `Usage Error: This project is configured to use <pkgmgr>`.

This happens when I run `npm i -g vercel@latest` in a project that is configured to use yarn or pnpm. This seems a little too strict because the global flag (-g) is not scoped to my current project.

Similarly, `vercel dev` will shell out to [`npm`](https://github.com/vercel/vercel/blob/7e0dd6f8087b574b09d796c5660e80df6e25afaa/packages/cli/src/util/dev/builder-cache.ts#L247-L251) or [`npx`](https://github.com/vercel/vercel/blob/7e0dd6f8087b574b09d796c5660e80df6e25afaa/packages/cli/src/util/dev/server.ts#L2221-L2235) and corepack being enabled could block it from working properly, even if the system has `npm` installed.

While I think the intention is good (catch accidental typos), I think this shouldn't be the default behavior because it is too restrictive. Perhaps making this strict mode be opt-in would help Corepack go stable (see #104).

aduh95:
Or it could spawn a prompt: `You are running a command the uses <pkgmng> in a directory configured to use <pkgmng>. Do you want to continue?`.

mehulkar:
I would also like to be able to run things like `yarn --version` or `npm --version` in a situation where corepack is enabled. This early `UsageError` prevents me from doing that. While debugging @nathanhammond and I noticed that [`config.json`](https://github.com/nodejs/corepack/blob/main/config.json) contains an allowlist of commands that are passed through: https://github.com/nodejs/corepack/blob/312d9ea6f107ef05c7e8955538b9abd9f857f8a6/sources/main.ts#L46-L55

I'm not sure if my case is a good use case for this config, but may be a good place to start?

styfle:
> Or it could spawn a prompt:

I think that would be even more annoying because then you need a way to pass `--yes` to corepack without it passing `--yes` to the underlying package manager. Maybe a warning would be sufficient. A hard block is just too much of a breaking change.

The only thing that might be safe to hard block is `<pkg> install`. So perhaps the allowlist needs to be inverted to a denylist so most commands can be allowed.

aduh95:
> I think a prompt would be annoying and cause more problems because you need a way to pass `--yes` to corepack without it passing `--yes` to the underlying package manager.

Using an env variable would be more practical than a `--yes` flag IMO.

[...]

PR #158, opened 2022-08-10
fix: add `npm view` to transparent commands

styfle:
- Related to https://github.com/nodejs/corepack/issues/157


aduh95 (approved):
Docs: https://docs.npmjs.com/cli/v8/commands/npm-view

ljharb:
it kind of seems like it'd be a more usable default to assume everything is transparent, and explicitly exclude the *opaque* ones.

arcanis:
> it kind of seems like it'd be a more usable default to assume everything is transparent, and explicitly exclude the _opaque_ ones.

Allowing new commands isn't a major change, whereas forbidding some is.

Generally I'm not sure how I feel about this change. While on surface it makes some sense, in practice it bypasses any registry configuration configured for the project (which can lead to potential security problems if the information you get is obtained from the public one instead of the private one). To avoid that one should call `yarn npm info` instead, which leverages the Yarn configuration.

There's an argument that this won't matter in most cases, but there's also the opposite argument: the cases where this matter are more likely to make this easy mistake unless the system prevents them to do so.

ljharb:
> Allowing new commands isn't a major change, whereas forbidding some is.

that is a very good point. another alternative is disallowing *all* commands unless they're explicitly marked as opaque or transparent - then it'd always be semver-minor to mark a new command.

aduh95:
@styfle What do you think of https://github.com/nodejs/corepack/pull/158#issuecomment-1215600771? Going back to my idea of providing an ENV variable to disable the usage error would let user/CLI tools that know what they're doing bypass the Corepack check while avoiding to introduce errors for the most common case. (So you would do something like `COREPACK_IGNORE_PACKAGE_JSON=1 npm view …` or `COREPACK_IGNORE_PACKAGE_JSON=1 npm install -g …`)

arcanis:
Fwiw I agree an environment variable would be reasonable.

Bikeshedding the name, I'd suggest `COREPACK_ENABLE_STRICT_CHECK`, which would default to `1`; it would match other similar settings, avoid the double negation, and is less ambiguous than just referencing the package.json.

styfle:
How about `COREPACK_ENABLE_STRICT=1` to enable the error and `COREPACK_ENABLE_STRICT=0` to disable it?

aduh95:
@styfle  do you still want this merged now that https://github.com/nodejs/corepack/pull/167 has landed?

styfle:
> While on surface it makes some sense, in practice it bypasses any registry configuration configured for the project

Given this, I think we should close and rely on the new `COREPACK_ENABLE_STRICT=0` env var.

I'm also curious if we can make `-g` or `--global` transparent by default but perhaps I'm overlooking something in those cases too.

PR #159, opened 2022-08-10
fix: add `npm bin` to transparent commands

styfle:
- Related to https://github.com/nodejs/corepack/issues/157

### DIFF
--- a/config.json
+++ b/config.json
@@ -6,6 +6,10 @@
         "commands": [
           [
             "npm",
+            "bin"
+          ],
+          [
+            "npm",
             "init"
           ],
           [


aduh95:
Sorry for my ignorance, what's the use-case for using `npm bin` in a non-npm package?

Docs: https://docs.npmjs.com/cli/v8/commands/npm-bin

arcanis:
I agree, I don't see how `npm bin` is relevant to non-npm projects (except maybe when the `-g` flag is set, but this PR doesn't check that).

styfle:
> what's the use-case for using npm bin in a non-npm package?

Its useful to append to the `PATH` so that you can run scripts like `tsc` instead of `./node_modules/.bin/tsc`.

https://github.com/vercel/vercel/blob/e34858d0828d97a7ccdedc36ff739cd051e20101/packages/next/src/index.ts#L429-L430

ljharb:
Note that npm 9 is removing `npm bin`

styfle:
@ljharb Where do you see that? Is there a roadmap I missed?

ljharb:
@styfle https://github.com/npm/statusboard/issues/537

merceyz:
> > what's the use-case for using npm bin in a non-npm package?
> 
> Its useful to append to the `PATH` so that you can run scripts like `tsc` instead of `./node_modules/.bin/tsc`.
> 
> https://github.com/vercel/vercel/blob/e34858d0828d97a7ccdedc36ff739cd051e20101/packages/next/src/index.ts#L429-L430

That's assuming the package manager used by the project implements the same behaviour as `npm` which isn't guaranteed.

ljharb:
It's also worth noting that that use case is wildly insecure and dangerous, because of the potential for a package to silently shadow a system command.

PR #167, opened 2022-08-24
feat: add `COREPACK_ENABLE_STRICT` env variable

aduh95:
Fixes: https://github.com/nodejs/corepack/issues/157

### DIFF
--- a/README.md
+++ b/README.md
@@ -111,11 +111,17 @@ This command will retrieve the given package manager from the specified archive 
 
 ## Environment Variables
 
-- `COREPACK_ENABLE_NETWORK` can be set to `0` to prevent Corepack from accessing the network (in which case you'll be responsible for hydrating the package manager versions that will be required for the projects you'll run, using `corepack hydrate`).
-
 - `COREPACK_DEFAULT_TO_LATEST` can be set to `0` in order to instruct Corepack
   not to lookup on the remote registry for the latest version of the selected
   package manager.
+
+- `COREPACK_ENABLE_NETWORK` can be set to `0` to prevent Corepack from accessing
+  the network (in which case you'll be responsible for hydrating the package
+  manager versions that will be required for the projects you'll run, using
+  `corepack hydrate`).
+
+- `COREPACK_ENABLE_STRICT` can be set to `0` to prevent Corepack from checking
+  if the package manager corresponds to the one defined for the current project.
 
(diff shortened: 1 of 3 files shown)


styfle (approved):
Thanks!

README.md
```diff
@@ -111,12 +111,18 @@ This command will retrieve the given package manager from the specified archive
...
 - `COREPACK_DEFAULT_TO_LATEST` can be set to `0` in order to instruct Corepack
   not to lookup on the remote registry for the latest version of the selected
   package manager.
 
+- `COREPACK_ENABLE_NETWORK` can be set to `0` to prevent Corepack from accessing
+  the network (in which case you'll be responsible for hydrating the package
+  manager versions that will be required for the projects you'll run, using
+  `corepack hydrate`).
+
+- `COREPACK_ENABLE_STRICT` can be set to `0` to prevent Corepack from checking
```
nathanhammond, line 123: It's a _bit_ weird that the name is `COREPACK_ENABLE_STRICT` but the default is `true`. You only use this environment variable when you want to _disable_ strict checking.

aduh95: It follows the same pattern as `COREPACK_ENABLE_NETWORK`. Do you have another name suggestion?

styfle: I prefer env vars and booleans alike are always "enable" regardless of default value so you don't have to think about double natives of disabling the "disable" variable.

ljharb: Separately tho, it's always weird and confusing when a boolean option has different behavior between "absent" and "false".

sources/specUtils.ts
```diff
@@ -44,6 +44,8 @@ export async function findProjectSpec(initialCwd: string, locator: Locator, {tra
   // A locator is a valid descriptor (but not the other way around)
   const fallbackLocator = {name: locator.name, range: locator.reference};
 
+  if (process.env.COREPACK_ENABLE_STRICT === `0`) return fallbackLocator;
```
arcanis, line 47: Nit: might not be needed in this PR, but it'd make sense to move the "parsing" logic in a function so we could be sure the same one is used for enableStrict and enableNetwork

arcanis (approved):

README.md
nathanhammond: I don't have another naming suggestion given that this follows an existing pattern. I'm not trying to block or delay this with my comment.

PR #168, opened 2022-08-24
chore: reduce test flakiness

aduh95:
IIUC `nock` needs to run in the same process as being tested, and since most tests spawn their own child process, the previous setup was mostly ignored. Moving it to the child process seems to improve things.


.github/workflows/ci.yml
```diff
@@ -78,3 +78,5 @@ jobs:
       - run: corepack yarn build # We need the stubs to run the tests
 
       - run: corepack yarn test
+        env:
+          - NOCK_ENV: replay
```
aduh95, line 82:
```suggestion
          NOCK_ENV: replay
```

.yarnrc.yml
```diff
@@ -0,0 +1 @@
+nodeLinker: node-modules
```
aduh95, line 1: We need this for the child process to be able to resolve `nock` (I think the alternative would be to bundle `recordRequests`, however changing the linker seemed easier to implement).

aduh95:
@arcanis do you have an opinion on using `node-modules` linker? Or maybe there's a way of avoid that that I'm missing? Otherwise I'd like to marge this.

.yarnrc.yml
merceyz: The child process should be able to resolve it without changing the linker, what error are you getting?

aduh95: Yeah that seems to be correct, not sure what I did to see the error (I probably ran `node` instead of `yarn node` in my manual tests) but yeah the tests results are the same regardless of the linker. Let me revert that.

PR #197, opened 2022-10-20
feat: when strict checking is off, treat like transparent

### LINKED ISSUE #195: Use the package manager version specified by the project, but still allow other package manager invocations
At time of writing, at 0.14.0 there is `COREPACK_ENABLE_STRICT=0` , which disables this check completely and will always use the fallback version of package managers. However, there should be a middle ground, to use the package manager version specified by the project, but does not throw error when other package manager is invoked.

Use case example:

- globally, I prepared & activated pnpm@7.13.4 and yarn@1.22.19
- there is a project, managed using pnpm@6.33.0 (specified in the "packageManager" field in package.json). However there are some external dependencies with pre/post-install scripts that calls "yarn build".

With the default configuration of Corepack, I cannot bootstrap the project at all, i.e. I cannot successfully run `pnpm install`. This is because invocation of yarn inside that repo will fail due to "Usage error", i.e. https://github.com/nodejs/corepack/issues/157)

However, with `COREPACK_ENABLE_STRICT=0`, I still cannot bootstrap the project as expected, because now `pnpm install` will use the fallback/global pnpm (7.13.4) instead of the version specified by the repo (6.33.0)

Proposal:

When  I set `COREPACK_ENABLE_STRICT=0`
- running `pnpm` in that project should still use the version specified by the repo (6.33.0) instead of the fallback/global version.
- running `yarn` in that project should use the global/fallback version

What do you think?

kenrick95:
Fixes #195 

Relevant to #167 and #157


aduh95 (commented):
Would it make sense to add another value to `COREPACK_ENABLE_STRICT` to keep a way to disable `package.json` parsing? Maybe `COREPACK_ENABLE_STRICT=2` or something? It doesn't have to block this, as it seems the implementation covers the more frequent use case, but I feel like Corepack should give a way to skip all interactions with `package.json`.

tests/main.test.ts
```diff
@@ -270,6 +270,11 @@ it(`should refuse to run a different package manager within a configured project
     process.env.COREPACK_ENABLE_STRICT = `0`;
 
     try {
+      await expect(runCli(cwd, [`yarn`, `--version`])).resolves.toMatchObject({
+        stdout: `1.0.0\n`,
+        stderr: ``,
+        exitCode: 0,
+      });
       await expect(runCli(cwd, [`pnpm`, `--version`])).resolves.toMatchObject({
```
kenrick95, line 278: I'm quite lost why tests failed in CI. I'm able to successfully run the tests in my local machine (and it all passes) 😅

kenrick95:
> Would it make sense to add another value to `COREPACK_ENABLE_STRICT` to keep a way to disable `package.json` parsing? Maybe `COREPACK_ENABLE_STRICT=2` or something? It doesn't have to block this, as it seems the implementation covers the more frequent use case, but I feel like Corepack should give a way to skip all interactions with `package.json`.

Yeah I think that can be brought up in a separate issue. Although on this matter, my view is that, that behavior removes Corepack's main feature of choosing project-defined package manager

aduh95:
> > Would it make sense to add another value to `COREPACK_ENABLE_STRICT` to keep a way to disable `package.json` parsing? Maybe `COREPACK_ENABLE_STRICT=2` or something? It doesn't have to block this, as it seems the implementation covers the more frequent use case, but I feel like Corepack should give a way to skip all interactions with `package.json`.
> 
> Yeah I think that can be brought up in a separate issue. Although on this matter, my view is that, that behavior removes Corepack's main feature of choosing project-defined package manager, and that makes Corepack yet another npm/yarn/pnpm version manager 😅

I can see it being useful as a debug feature, not for production – while I could see why the behavior implemented in this PR would be useful in production, it's two different use cases.

kenrick95:
> > > Would it make sense to add another value to `COREPACK_ENABLE_STRICT` to keep a way to disable `package.json` parsing? Maybe `COREPACK_ENABLE_STRICT=2` or something? It doesn't have to block this, as it seems the implementation covers the more frequent use case, but I feel like Corepack should give a way to skip all interactions with `package.json`.
> > 
> > 
> > Yeah I think that can be brought up in a separate issue. Although on this matter, my view is that, that behavior removes Corepack's main feature of choosing project-defined package manager, and that makes Corepack yet another npm/yarn/pnpm version manager 😅
> 
> I can see it being useful as a debug feature, not for production – while I could see why the behavior implemented in this PR would be useful in production, it's two different use cases IMHO.

I see. Though I feel it would be better as a separate flag, like `COREPACK_IGNORE_PROJECT_SPEC`? `COREPACK_ENABLE_STRICT=2` gives the impression of Corepack being extra strict, but in reality, it is not 😅

arcanis:
I tend to agree, if the feature was still there it'd make more sense as a separate flag imo (`COREPACK_ENABLE_PROJECT_SPEC=0`, maybe, by consistency with the other flags?)

aduh95 (approved):

tests/main.test.ts
aduh95: Yeah I had to fight with the CI a bit, it's finally green now 😅

Issue #198, opened 2022-10-20
corepack hydrate fails if $HOME/.cache/node/corepack does not exist


Gamadril:
Platform: MacOS x64
NodeJS version: v16.18.0
Corepack version: 0.14.1

Steps to reproduce:
- download nodejs 16.18.0 dist (https://nodejs.org/dist/v16.18.0/) (darwin-x64 in my case)
- cd into it
- run "export PATH=./bin:$PATH" to override globally installed nodejs version
- run "corepack enable"
- run "corepack prepare yarn@3.2.4 -o"
- empty $HOME/.cache folder
- run "corepack hydrate corepack.tgz"

Result:
```
Hydrating yarn@3.2.4...
Cwd Error: ENOENT: Cannot cd into '/Users/xxx/.cache/node/corepack'
    at /Users/xxx/Downloads/node-v16.18.0 /lib/node_modules/corepack/dist/corepack.js:6725:12
    at FSReqCallback.oncomplete (node:fs:202:21)
```

If I manually create $HOME/.cache/node/corepack folder, the hydrate command works.

Expected:
corepack creates the folder structure automatically if it does not exist.

aduh95:
Do you want to send a PR to fix that?

Gamadril:
I will try, however I cannot event build it.
Followed the instruction from README:

```
    - Clone this repository
    - Run yarn build (no need for yarn install)
    - The dist/ directory now contains the corepack build and the shims
    - Call node ./dist/corepack --help and behold
```

Cloned the repo.
"yarn build" fails:
```
yarn run v1.22.19
$ rm -rf dist shims && webpack && ts-node ./mkshims.ts
/bin/sh: webpack: command not found
error Command failed with exit code 127.
```

run "yarn" to install dependencies first, "yarn build" still fails:

```
ERROR in /Users/xxx/Projects/node/corepack/sources/commands/Hydrate.ts
./sources/commands/Hydrate.ts 40:29-35
[tsl] ERROR in /Users/xxx/Projects/node/corepack/sources/commands/Hydrate.ts(40,30)
      TS2339: Property 'header' does not exist on type 'ReadEntry'.
ts-loader-default_06006a6d86bad3e7
 @ ./sources/main.ts 5:0-52 85:21-35
 @ ./sources/_entryPoint.ts 1:0-33 3:0-19 6:4-11

webpack 5.74.0 compiled with 1 error and 3 warnings in 8154 ms
error Command failed with exit code 1.
```

[...]

PR #200, opened 2022-10-20
Checks if the new corepack home folder exists and creates it if necessary (Fix for issue #198)

Gamadril:


merceyz (changes requested):
Could you add a test and update the PR title to describe what is going on? Seeing `Fix for issue #198` in the commit history doesn't really say much about what happened.

Gamadril:
Prepare is not Hydrate... where actually the error happens...

aduh95:
> Prepare is not Hydrate... where actually the error happens...

Yeah we would need one before `tar.x` as well: https://github.com/nodejs/corepack/blob/b77ed59321ff055de7cb23e6ca75d26c35916ee8/sources/commands/Hydrate.ts#L66

sources/folderUtils.ts
```diff
@@ -25,6 +25,7 @@ export function getInstallFolder() {
       mkdirSync(newCorepackDefaultHome, {recursive: true});
       renameSync(oldCorepackDefaultHome, newCorepackDefaultHome);
     }
+
```
aduh95, line 28:
```suggestion
```

sources/commands/Hydrate.ts
```diff
@@ -1,5 +1,6 @@
 import {Command, Option, UsageError} from 'clipanion';
 import path                          from 'path';
+import {existsSync, mkdirSync}       from 'fs';
```
aduh95, line 3:
```suggestion
import path                          from 'path';
import {mkdir}                       from 'fs/promises';
```

sources/commands/Hydrate.ts
```diff
@@ -36,6 +37,9 @@ export class HydrateCommand extends Command<Context> {
 
     const {default: tar} = await import(/* webpackMode: 'eager' */ `tar`);
 
+    if (!existsSync(installFolder))
+      mkdirSync(installFolder, {recursive: true});
```
aduh95, line 41: I don't think we gain anything from the `existSync` check, it's already built-in `mkdir` when `recursive` is `true`.
```suggestion
  // Recreate the folder in case it was deleted somewhere else:
  await mkdir(installFolder, {recursive: true});
```

This needs to be moved below, in the loop, in case the directory is deleted while `corepack` is running, we want to re-create it just before the `tar.x` command runs.

aduh95 (commented):
The `mkdir` call needs to happen just before `tar.c` in `Prepare.ts` and `tar.x` in `Hydrate.ts`.

aduh95:
@Gamadril Any chance you'll have time to address the remaining suggestions and add some tests so we can land this?

Gamadril:
hmm, seems to be more complicated than expected. will have to investigate more time...

sources/commands/Hydrate.ts
```diff
@@ -37,8 +37,8 @@ export class HydrateCommand extends Command<Context> {
 
     const {default: tar} = await import(/* webpackMode: 'eager' */ `tar`);
 
-    if (!existsSync(installFolder))
-      mkdirSync(installFolder, {recursive: true});
+    // Recreate the folder in case it was deleted somewhere else:
+    await mkdir(installFolder, {recursive: true});
```
aduh95, line 41: I don't know if you see my previous comment, this needs to be moved below: the `mkdir` call needs to happen just before `tar.x`, not `tar.t`.

Gamadril: sorry, my mistake

aduh95 (approved):
LGTM, would you like to add a fix and a test for the `prepare` command as well?

tests/main.test.ts
```diff
@@ -404,6 +404,44 @@ it(`should support hydrating package managers from cached archives`, async () =>
   });
 });
 
+it(`should support hydrating package managers if cache folder does not exist`, async () => {
```
aduh95, line 407:
```suggestion
it(`should support hydrating package managers if cache folder was removed`, async () => {
```

PR #186, opened 2022-09-24
feat: add support for configurable registries and applicable authentication options

micsco:
This adds support per #66 

Resolves https://github.com/nodejs/corepack/issues/66

PR #219, opened 2022-12-30
fix: download registry

xiaoxiangmoe:
Fix download registry.

---

**Enviroment**:
* corepack 0.15.2
* node 19.3.0

**Bug description**: 

In China, access to `https://registry.npmjs.org`,`https://registry.yarnpkg.com/` and `https://repo.yarnpkg.com/` is unstable. We should use `https://registry.npmmirror.com` instead.

We can modify our `/etc/host` file add a line `127.0.0.1 registry.npmjs.org` to simulate npm being unreachable.

Then run `COREPACK_NPM_REGISTRY="https://registry.npmmirror.com" corepack prepare pnpm@7.17.0 --activate`. It will try to download `https://registry.npmjs.org/pnpm/-/pnpm-7.17.0.tgz`. COREPACK_NPM_REGISTRY is not working.


xiaoxiangmoe:
@micsco @aduh95 This PR is releated to https://github.com/nodejs/corepack/pull/186

xiaoxiangmoe:
This works for npm and pnpm

definitions.npm.ranges['*'].url:
https://github.com/nodejs/corepack/blob/f84cfcb00ffb985d44b6aa0b563b2b4056a8f0d0/config.json#L22
definitions.pnpm.ranges['>=6.0.0'].url:
https://github.com/nodejs/corepack/blob/f84cfcb00ffb985d44b6aa0b563b2b4056a8f0d0/config.json#L68

---

But it seems not work for yarn.
definitions.yarn.ranges['<2.0.0'].url:
https://github.com/nodejs/corepack/blob/f84cfcb00ffb985d44b6aa0b563b2b4056a8f0d0/config.json#L97
definitions.yarn.ranges['>=2.0.0'].url:
https://github.com/nodejs/corepack/blob/f84cfcb00ffb985d44b6aa0b563b2b4056a8f0d0/config.json#L109

sources/corepackUtils.ts
```diff
@@ -104,7 +104,7 @@ export async function installVersion(installTarget: string, locator: Locator, {s
     return installFolder;
   }
 
-  const url = spec.url.replace(`{}`, version);
+  const url = spec.url.replace(`{}`, version).replace(npmRegistryUtils.DEFAULT_NPM_REGISTRY_URL, process.env.COREPACK_NPM_REGISTRY || npmRegistryUtils.DEFAULT_NPM_REGISTRY_URL);
```
aduh95, line 107:
```suggestion
  const defaultNpmRegistryURL = spec.url.replace(`{}`, version)
  const url = process.env.COREPACK_NPM_REGISTRY ?
    defaultNpmRegistryURL.replace(npmRegistryUtils.DEFAULT_NPM_REGISTRY_URL, () => process.env.COREPACK_NPM_REGISTRY) :
    defaultNpmRegistryURL;
```

aduh95:
> Should we change the yarn download link in `config.json` to download from npm registry?

AFAIK Yarn 2+ versions are no longer published on the npm registry[^a] (the last 2+ version published was 2.4.3 published in September 2021), using an npm mirror wouldn't work. I'm not sure what would be the best way to make that customizable, adding a `COREPACK_YARN_REGISTRY` would not be a great idea, as if more package managers decide to use a custom registry it's going to be unmaintainable.

[^a]: https://www.npmjs.com/package/yarn?activeTab=versions

arcanis:
Perhaps we could have a generic `COREPACK_DNS_OVERRIDE=registry.something.com/registry.somethingelse.com` which would replace all calls to the first domain by calls to the second? This would work with both npm and custom registries.

xiaoxiangmoe:
@arcanis @aduh95 Do you have any idea about how to use COREPACK_DNS_OVERRIDE to config both npm registery and yarn registery?

arcanis:
Perhaps, to ensure the syntax is somewhat familiar, we can make it a JSON object? For example:

```
COREPACK_DNS_OVERRIDE='{
  "registry.yarnpkg.com": "mirror.something.com",
  "registry.npmjs.com": "mirror.somethingelse.com"
}' yarn
```

Main problem are the newlines which, while they seem to not be an actual problem when use in env variables *per se*, may cause wonky behaviours on scripts making incorrect assumptions about the result of `$ env`.

Alternatively, perhaps we could have some `corepack` settings in the closest `package.json`, since we already know its location to extract the package manager to use.

aduh95:
> Main problem are the newlines

Worth noting that JSON does not require using new lines, so if the new lines are a problem, a simple workaround is to not use them.

> Alternatively, perhaps we could have some `corepack` settings in the closest `package.json`, since we already know its location to extract the package manager to use.

I don't think it's a good idea, a DNS override is not something that's defined on a project basis, but rather as a system level, so env variable would make the most sense. Or am I missing something?

xiaoxiangmoe:
@aduh95 Do you have any idea about new env variable for this featurer to replacing `COREPACK_DNS_OVERRIDE`?

sources/corepackUtils.ts
```diff
@@ -104,7 +104,10 @@ export async function installVersion(installTarget: string, locator: Locator, {s
     return installFolder;
   }
 
-  const url = spec.url.replace(`{}`, version);
+  const defaultNpmRegistryURL = spec.url.replace(`{}`, version);
+  const url = process.env.COREPACK_NPM_REGISTRY ?
+    defaultNpmRegistryURL.replace(npmRegistryUtils.DEFAULT_NPM_REGISTRY_URL, () => process.env.COREPACK_NPM_REGISTRY) :
```
aduh95, line 109:
```suggestion
    defaultNpmRegistryURL.replace(
      npmRegistryUtils.DEFAULT_NPM_REGISTRY_URL,
      () => process.env.COREPACK_NPM_REGISTRY!,
    ) :
```

aduh95:
@xiaoxiangmoe if you want to make `COREPACK_DNS_OVERRIDE` happen, that'd be awesome! In the mean time, let's merge this and release so it can already unblock users of npm and pnpm.

PR #222, opened 2023-01-01
refactor: simplify `v8-compile-cache` setup

merceyz:
After the changes in https://github.com/nodejs/corepack/pull/97 we no longer need a path to `v8-compile-cache` so we can just import it normally.

### DIFF
--- a/mkshims.ts
+++ b/mkshims.ts
@@ -6,9 +6,6 @@ import {Engine}                     from './sources/Engine';
 import {SupportedPackageManagerSet} from './sources/types';
 
 function shouldGenerateShim(name: string) {
-  if (name.startsWith(`vendors`))
-    return false;
-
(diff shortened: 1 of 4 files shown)


mkshims.ts
```diff
@@ -6,11 +6,9 @@ import {Engine}                     from './sources/Engine';
...
 
 function shouldGenerateShim(name: string) {
-  if (name === 'vcc.js') {
+  if (name.startsWith(`vendors`))
     return false;
-  } else if (name.startsWith('vendors')) {
-    return false;
-  }
+
   return true;
```
aduh95, line 12: nit: Not a fan of the `if(condition) return false else return true` pattern
```suggestion
  return !name.startsWith(`vendors`);
```

aduh95 (approved):

serroopoint:
#

Issue #130, opened 2022-06-17
File name too long for ustar when creating a tarball of the nodejs installation


cemkeylan:
### Version

v18.4.0

### Platform

`Linux dell-carbs 5.15.10 #7 SMP Tue Jan 18 15:17:48 +03 2022 x86_64 Carbs Linux`

### What steps will reproduce the bug?

Create a tarball of the installed files using a standards-compliant `tar` or `pax` program.

``` sh
./configure \
    --prefix=/usr \
    --ninja

ninja -C out/Release
tools/install.py install pkg /usr

tar -czf ninja.tar.gz pkg
```

or

```
pax -wz pkg > ninja.tar.gz
```

### How often does it reproduce? Is there a required condition?

As long as the program follows the `ustar` format specifications, which limits file names to be 100 characters or less, it can be reproduced everytime. OpenBSD's implementation of `tar` is an example for this.

### What is the expected behavior?

Creating a tarball without any errors.

### What do you see instead?

```
tar: File name too long for ustar usr/lib/node_modules/corepack/dist/vendors-_yarn_berry_cache_proxy-agent-npm-5_0_0-41772f4b01-9_zip_node_modules_proxy-agent_index_js.js
```

cemkeylan:
As a temporary measure I have renamed the file to be shorter and replaced the name in `corepack.js' which seems to be working fine when I enable corepack and use `yarn`

PR #224, opened 2023-01-03
build: enable production mode

merceyz:

Corepack is built in Webpack's development mode which causes Webpack to output files with long filenames.

Fixes https://github.com/nodejs/corepack/issues/130

**How did you fix it?**

Change it to build in production mode but disable minimizing so the code is still somewhat readable.

The bundle size also improved a bit:
```diff
$ du -s dist
- 2434    dist
+ 2170    dist
```

### DIFF
--- a/mkshims.ts
+++ b/mkshims.ts
@@ -6,7 +6,7 @@ import {Engine}                     from './sources/Engine';
 import {SupportedPackageManagerSet} from './sources/types';
 
 function shouldGenerateShim(name: string) {
-  return !name.startsWith(`vendors`);
+  return name !== `chunks`;
 }
 
 const engine = new Engine();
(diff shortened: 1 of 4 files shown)


styfle (commented):
Removing terser should probably be a separate PR since its not related to `production` mode.

I think its acceptable to merge the `production` change because that fixes a bug. But removing terser is more of a preference and might require further discussion.

merceyz:
> But removing terser is more of a preference and might require further discussion.

"Removing" doesn't seem quite right; in `development` mode it didn't do anything and this PR preserves that by disabling it with `minimize: false`, therefore the `new TerserPlugin` code was unnecessary so I removed it.

arcanis (approved):
LGTM; I probably misunderstood whether `development` mode would respect or not explicit optimization settings.

PR #227, opened 2023-01-04
chore: add `engines.node` to `package.json`

merceyz:

[v0.11.0](https://github.com/nodejs/corepack/releases/tag/v0.11.0) (specifically https://github.com/nodejs/corepack/pull/100) dropped support for Node.js v12 and raised the bar to Node.js v14.14.0 due to https://github.com/nodejs/corepack/blob/82a6715702ac49687aa8b731efb5971bcdeac0a3/sources/fsUtils.ts#L4 (https://nodejs.org/dist/latest-v18.x/docs/api/fs.html#fspromisesrmpath-options) but this wasn't written down as far as I can tell.

Addresses https://github.com/nodejs/corepack/pull/100#pullrequestreview-945341129

**How did you fix it?**

Added `engines.node` to `package.json` to document the minimum version.

### DIFF
--- a/tsconfig.json
+++ b/tsconfig.json
@@ -7,12 +7,12 @@
     "moduleResolution": "node",
     "noEmit": true,
     "forceConsistentCasingInFileNames": true,
-    "lib": ["dom", "es2017", "esnext.asynciterable"],
+    "lib": ["ES2020"],
     "module": "commonjs",
     "resolveJsonModule": true,
     "skipLibCheck": true,
     "strict": true,
-    "target": "es2017"
+    "target": "ES2020"
   },
   "ts-node": {
     "transpileOnly": true
--- a/webpack.config.js
+++ b/webpack.config.js
@@ -22,7 +22,7 @@ module.exports = {
         loader: `ts-loader`,
         options: {
           compilerOptions: {
-            module: `es2020`,
+            module: `ES2020`,
             noEmit: false,
           },
         },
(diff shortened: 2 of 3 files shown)


tsconfig.json
```diff
@@ -7,12 +7,12 @@
...
     "noEmit": true,
     "forceConsistentCasingInFileNames": true,
-    "lib": ["dom", "es2017", "esnext.asynciterable"],
+    "lib": ["ES2020"],
     "module": "commonjs",
     "resolveJsonModule": true,
     "skipLibCheck": true,
     "strict": true,
-    "target": "es2017"
+    "target": "ES2020"
```
merceyz, line 15: According to https://node.green/#ES2020 Node.js v14.14 doesn't fully support ES2020 but we're not using `spread parameters after optional chaining` so that's fine, if we ever do use it the CI will catch it assuming v14 doesn't update to a version of v8 that supports it.
In a future PR I'll ensure `spread parameters after optional chaining` is transpiled.

arcanis (approved):

webpack.config.js
```diff
@@ -22,7 +22,7 @@ module.exports = {
         loader: `ts-loader`,
         options: {
           compilerOptions: {
-            module: `es2020`,
+            module: `ES2020`,
```
styfle, line 25: I wonder if this should read from tsconfig.json here to keep them in sync?

merceyz: That's a good idea however my next PR will replace webpack with esbuild so they shouldn't have time to go out of sync.

aduh95 (approved):

PR #230, opened 2023-01-11
chore: Use setup-node action to cache dependencies

jongwooo:
Signed-off-by: jongwooo <jongwooo.han@gmail.com>

## Description

Updated workflows to cache dependencies using [actions/setup-node](https://github.com/actions/setup-node#caching-global-packages-data). `setup-node@v3` or newer has caching **built-in**.

### AS-IS

```yaml
- uses: actions/setup-node@v3
  with:
    node-version: lts/*

- name: Get yarn cache directory path
  id: yarn-cache-dir-path
  run: echo "::set-output name=dir::$(corepack yarn config get cacheFolder)"

- uses: actions/cache@v3
  with:
    path: ${{steps.yarn-cache-dir-path.outputs.dir}}
    key: ${{runner.os}}-yarn-${{hashFiles('**/yarn.lock')}}
    restore-keys: |
      ${{runner.os}}-yarn-
```

### TO-BE

```yml
- uses: actions/setup-node@v3
  with:
    node-version: lts/*
    cache: yarn
```

> It’s literally a one line change to pass the `cache: yarn` input parameter.

## References

- [https://docs.github.com/en/actions/using-workflows/caching-dependencies-to-speed-up-workflows](https://docs.github.com/en/actions/using-workflows/caching-dependencies-to-speed-up-workflows)

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -20,17 +20,7 @@ jobs:
       - uses: actions/setup-node@v3
         with:
           node-version: lts/*
-
-      - name: Get the Yarn cache directory path
-        id: yarn-cache-dir-path
-        run: echo "::set-output name=dir::$(corepack yarn config get cacheFolder)"
-
-      - uses: actions/cache@v3
-        with:
-          path: ${{steps.yarn-cache-dir-path.outputs.dir}}
-          key: ${{runner.os}}-yarn-${{hashFiles('**/yarn.lock')}}
-          restore-keys: |
-            ${{runner.os}}-yarn-
+          cache: yarn
 
       - run: corepack yarn install --immutable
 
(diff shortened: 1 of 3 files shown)


merceyz:
The caching done by `actions/setup-node` doesn't have any `restore-keys` set so this will cause worse performance when there isn't an exact match (https://github.com/actions/setup-node/issues/325).

`actions/setup-node` also doesn't prefix its Yarn commands with `corepack` so the CI would need to run `corepack enable` before `actions/setup-node` runs.

jongwooo:
> `actions/setup-node` also doesn't prefix its Yarn commands with `corepack` so the CI would need to run `corepack enable` before `actions/setup-node` runs.

@merceyz I didn't know about this part. Thank you for informing me! Is it okay for me to close this pull request if I may?

styfle (commented):
I've had a lot of problems in the past when using the default cache feature with yarn.

See these open issues:

- https://github.com/actions/cache/issues/810
- https://github.com/Azure/azure-sdk-for-js/issues/22321

merceyz:
That was an issue with Yarn classic, haven't seen the same issue in modern versions.

> Is it okay for me to close this pull request if I may?

Of course, you can close PRs for any reason.

PR #239, opened 2023-02-19
Add `"exports"` to the `package.json` to make self-referencing import…

aduh95:
…s work

I'm getting the following error, probably due to the switch to ESBuild (Webpack used to embed the JSON info in the JS file):

```
Internal Error: Cannot find module 'corepack/package.json'
Require stack:
- /Users/duhamean/Documents/node/deps/corepack/dist/corepack.js
Require stack:
- /Users/duhamean/Documents/node/deps/corepack/dist/corepack.js
    at Module._resolveFilename (node:internal/modules/cjs/loader:1060:15)
    at Function.resolve (node:internal/modules/helpers:118:19)
    at runVersion (…/node/deps/corepack/dist/corepack.js:44173:68)
```

This should fix it for me, but I'm unsure about others, maybe a better fix would be to embed the version string in the JS file? //cc @merceyz

### DIFF
--- a/tests/_runCli.ts
+++ b/tests/_runCli.ts
@@ -1,5 +1,6 @@
 import {PortablePath, npath} from '@yarnpkg/fslib';
 import {spawn}               from 'child_process';
+import {dirname, join}       from 'path';
 
 export async function runCli(cwd: PortablePath, argv: Array<string>): Promise<{exitCode: number | null, stdout: string, stderr: string}> {
   const out: Array<Buffer> = [];
@@ -8,7 +9,7 @@ export async function runCli(cwd: PortablePath, argv: Array<string>): Promise<{e
   return new Promise((resolve, reject) => {
     if (process.env.RUN_CLI_ID)
       (process.env.RUN_CLI_ID as any)++;
-    const child = spawn(process.execPath, [`--no-warnings`, `-r`, require.resolve(`./recordRequests.js`), require.resolve(`corepack/dist/corepack.js`), ...argv], {
+    const child = spawn(process.execPath, [`--no-warnings`, `-r`, require.resolve(`./recordRequests.js`), join(dirname(require.resolve(`corepack/package.json`)), `dist/corepack.js`), ...argv], {
       cwd: npath.fromPortablePath(cwd),
       env: process.env,
       stdio: `pipe`,
--- a/tests/_runCli.ts
+++ b/tests/_runCli.ts
@@ -1,6 +1,5 @@
 import {PortablePath, npath} from '@yarnpkg/fslib';
 import {spawn}               from 'child_process';
-import {dirname, join}       from 'path';
 
 export async function runCli(cwd: PortablePath, argv: Array<string>): Promise<{exitCode: number | null, stdout: string, stderr: string}> {
   const out: Array<Buffer> = [];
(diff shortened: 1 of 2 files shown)


merceyz:
I'm fine with this, it works when distributed by Node.js since it's located in a `node_modules` folder but seems that isn't true before its packaged.

> probably due to the switch to ESBuild (Webpack used to embed the JSON info in the JS file)

It should have been a problem before that as well due to [this line](https://github.com/nodejs/corepack/commit/15ceb832a34a223efbe3d3f9cb792d9101a7022a#diff-c91b034b85c9610409bf66de5d28560a06f1285ee1962f238fac2c3652a15e88L55) but I guess you didn't hit that code path, with the move to esbuild I changed [this hack](https://github.com/nodejs/corepack/commit/15ceb832a34a223efbe3d3f9cb792d9101a7022a#diff-b0586ed3b0d38f97c6a608f1c1877a9154a009b56731a7b220e74d22ea9984cbR197) which you do run into.

> maybe a better fix would be to embed the version string in the JS file?

We (esbuild) already do that, this is used to get the path to the root of the package.
https://github.com/nodejs/corepack/blob/15ceb832a34a223efbe3d3f9cb792d9101a7022a/sources/commands/Enable.ts#L53-L55
https://github.com/nodejs/corepack/blob/15ceb832a34a223efbe3d3f9cb792d9101a7022a/sources/corepackUtils.ts#L197

ljharb:
Note that adding exports is a breaking change.

aduh95:
How does this work when Corepack is installed with the Node.js installer? I would expect this hack not to work, 
unlike `eval(__dirname)` (at least I can see what it's doing).

> Note that adding exports is a breaking change.

Thanks for the remainder, that wouldn't be a big deal to release Corepack 0.17.0.

tests/_runCli.ts
```diff
@@ -8,7 +9,7 @@ export async function runCli(cwd: PortablePath, argv: Array<string>): Promise<{e
   return new Promise((resolve, reject) => {
     if (process.env.RUN_CLI_ID)
       (process.env.RUN_CLI_ID as any)++;
-    const child = spawn(process.execPath, [`--no-warnings`, `-r`, require.resolve(`./recordRequests.js`), require.resolve(`corepack/dist/corepack.js`), ...argv], {
+    const child = spawn(process.execPath, [`--no-warnings`, `-r`, require.resolve(`./recordRequests.js`), join(dirname(require.resolve(`corepack/package.json`)), `dist/corepack.js`), ...argv], {
```
merceyz, line 12:
```suggestion
    const child = spawn(process.execPath, [`--no-warnings`, `-r`, require.resolve(`./recordRequests.js`), require.resolve(`../dist/corepack.js`), ...argv], {
```

merceyz:
> How does this work when Corepack is installed with the Node.js installer?

Because in that case Corepack is located in `.../node_modules/corepack` (https://github.com/nodejs/corepack/pull/239#issuecomment-1435964852)

Adding `exports` makes it work regardless of the path to Corepack due to https://github.com/nodejs/node/blob/0e3b796cc543c8e7c151f02462912592f7debd2d/lib/internal/modules/cjs/loader.js#L528

PR #244, opened 2023-02-27
chore(docs): document how to use aliases instead of jumper binaries

aduh95:


README.md
```diff
@@ -131,12 +131,35 @@ on a project where the `packageManager` field references `pnpm`).
...
 is called without parameters). Note that the npm shims will not be installed
 unless explicitly requested, as npm is currently distributed with Node.js
 through other means.
 
+If the file system where the `corepack` binary is located is read-only, this
+command will fail. A workaround is to add the binaries as alias in your
+shell configuration file (e.g. in `~/.bash_aliases`):
+
+```sh
+alias yarn="corepack yarn"
```
merceyz, line 145:
```suggestion
alias yarn="corepack yarn"
alias yarnpkg="corepack yarnpkg"
```

README.md
```diff
@@ -131,12 +131,35 @@ on a project where the `packageManager` field references `pnpm`).
...
+alias pnpx="corepack pnpx"
+alias npm="corepack npm"
+alias npx="corepack npx"
+```
+
+On Windows PowerShell, you can add functions using the `$PROFILE` automatic
+variable:
+
+```powershell
+echo "function yarn { corepack yarn `$args }" >> $PROFILE
```
merceyz, line 156:
```suggestion
echo "function yarn { corepack yarn `$args }" >> $PROFILE
echo "function yarnpkg { corepack yarnpkg `$args }" >> $PROFILE
```

andersk:
This doesn’t work for third-party scripts that invoke `yarn` or `pnpm`, since shell aliases are generally only available to your interactive shell and not to scripts.

merceyz:
Yarn adds itself to `PATH` when running scripts so assuming those third party scripts were invoked through Yarn it will work.

aduh95:
> This doesn’t work for third-party scripts that invoke `yarn` or `pnpm`, since shell aliases are generally only available to your interactive shell and not to scripts.

I've been using this for a bit now, including aliasing `npm` and `npx`, and indeed there are some issues with some scripts that expects it to be an executable on the PATH, but I've always been able to work around it.

PR #258, opened 2023-04-18
chore: update supported Node.js versions

aduh95:
BREAKING CHANGE: remove support for Node.js 14.x

### DIFF
--- a/README.md
+++ b/README.md
@@ -15,7 +15,7 @@ use npm.
 
 ### Default Installs
 
-Corepack is distributed by default with Node.js 14.19.0 and 16.9.0, but is
+Corepack is distributed by default with all recent Node.js versions, but is
 opt-in for the time being. Run `corepack enable` to install the required shims.
 
 ### Manual Installs
--- a/README.md
+++ b/README.md
@@ -2,26 +2,20 @@
 
 Corepack is a zero-runtime-dependency Node.js script that acts as a bridge
 between Node.js projects and the package managers they are intended to be used
-with during development. In practical terms, **Corepack will let you use Yarn
-and pnpm without having to install them** - just like what currently happens
-with npm, which is shipped by Node.js by default.
-
-**Important:** At the moment, Corepack only covers Yarn and pnpm. Given that we
-have little control on the npm project, we prefer to focus on the Yarn and pnpm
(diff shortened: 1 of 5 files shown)


merceyz (changes requested):
It doesn't cost much, if anything, to support Node.js v14.14.0 so could you document why support for it is dropped and why specifically v16.20.0 was chosen as the minimum?

If this is approved the tsconfig should be updated to ES2021 and the esbuild target changed as well.
https://github.com/nodejs/corepack/blob/84e6e1d018c59b37c4742c7a9100eb21267e114d/tsconfig.json#L15
https://github.com/nodejs/corepack/blob/84e6e1d018c59b37c4742c7a9100eb21267e114d/tsconfig.json#L10
https://github.com/nodejs/corepack/blob/367a9fc92a8bdd7ae605440265949afda3093c0d/package.json#L54

aduh95:
> It doesn't cost much, if anything, to support Node.js v14.14.0

Pnpm v8 doesn't support Node.js v14.x anymore, so if we want to keep the pace, we have to drop it on CI, and if we drop it on CI, we should probably not document that we support it.

> why specifically v16.20.0 was chosen as the minimum?

It's the latest v16 version, I don't have the bandwidth to test all the versions, using the latest seems easier and safer anyway.

README.md
```diff
@@ -15,7 +15,7 @@ use npm.
 
 ### Default Installs
 
-Corepack is distributed by default with Node.js 14.19.0 and 16.9.0, but is
+Corepack is distributed by default with all recent Node.js versions, but is
```
styfle, line 18: I would prefer a specific version here because this is about the expectation of `corepack` cli being distributed with `node` in the current PATH.

Alternatively, we could link to the [official docs](https://nodejs.org/api/corepack.html) that mention the versions in the first section:

```
Added in: v16.9.0, v14.19.0
```

styfle:
> It doesn't cost much, if anything, to support Node.js v14.14.0 so could you document why support for it is dropped

Another reason to drop node@14 is that it reaches EOL on 2023-04-30 in a couple weeks so newer versions of Corepack won't be distributed with future node@14 because there will be none after that date.

merceyz (approved):

PR #261, opened 2023-04-28
chore: report error url on fetchUrlStream exception

tisonkun:
FOR ISSUE LOCATING...

### DIFF
--- a/sources/httpUtils.ts
+++ b/sources/httpUtils.ts
@@ -15,7 +15,7 @@ export async function fetchUrlStream(url: string, options: RequestOptions = {}) 
   return new Promise<IncomingMessage>((resolve, reject) => {
     const request = https.get(url, {...options, agent: proxyAgent}, response => {
       const statusCode = response.statusCode;
-      if (statusCode && statusCode >= 200 && statusCode < 300)
+      if (statusCode != null && statusCode >= 200 && statusCode < 300)
         return resolve(response);
 
       return reject(new Error(`Server answered with HTTP ${statusCode} when performing the request to ${url}; for troubleshooting help, see https://github.com/nodejs/corepack#troubleshooting`));
--- a/sources/httpUtils.ts
+++ b/sources/httpUtils.ts
@@ -15,10 +15,10 @@ export async function fetchUrlStream(url: string, options: RequestOptions = {}) 
   return new Promise<IncomingMessage>((resolve, reject) => {
     const request = https.get(url, {...options, agent: proxyAgent}, response => {
       const statusCode = response.statusCode;
-      if (!(statusCode >= 200 && statusCode < 300))
-        return reject(new Error(`Server answered with HTTP ${statusCode} when performing the request to ${url}; for troubleshooting help, see https://github.com/nodejs/corepack#troubleshooting`));
+      if (statusCode && statusCode >= 200 && statusCode < 300)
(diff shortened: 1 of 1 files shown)


aduh95:
Hello, can you fix the conflict please?

tisonkun:
@aduh95 Updated. It seems the other place has been fixed by another PR.

sources/httpUtils.ts
```diff
@@ -16,7 +16,7 @@ export async function fetchUrlStream(url: string, options: RequestOptions = {})
     const request = https.get(url, {...options, agent: proxyAgent}, response => {
       const statusCode = response.statusCode ?? 500;
```
aduh95, line 17: While we're at it, I think we should not use `500` for nullish statusCode, as it's likely going to confuse users more to have Corepack pretend it saw a 500 response code when it fact it didn't see anything.
```suggestion
      const statusCode = response.statusCode;
```

tisonkun: Thanks for your spot. Updated.

tisonkun:
I pushed a new commit to try fix typecheck.

sources/httpUtils.ts
```diff
@@ -14,11 +14,11 @@ export async function fetchUrlStream(url: string, options: RequestOptions = {})
 
   return new Promise<IncomingMessage>((resolve, reject) => {
     const request = https.get(url, {...options, agent: proxyAgent}, response => {
-      const statusCode = response.statusCode ?? 500;
-      if (!(statusCode >= 200 && statusCode < 300))
-        return reject(new Error(`Server answered with HTTP ${statusCode}`));
+      const statusCode = response.statusCode;
+      if (statusCode && statusCode >= 200 && statusCode < 300)
```
aduh95, line 18: We want an error if the status code is 0
```suggestion
      if (statusCode != null && statusCode >= 200 && statusCode < 300)
```

tisonkun: I'm afraid that typecheck complains with "TS2532: object is possibly 'undefined'."

aduh95: `!= null` catches both `null` and `undefined`, Ts shouldn’t complain here

tisonkun: Cool!

tisonkun: Updated.

sources/httpUtils.ts
```diff
@@ -14,11 +14,11 @@ export async function fetchUrlStream(url: string, options: RequestOptions = {})
 
   return new Promise<IncomingMessage>((resolve, reject) => {
     const request = https.get(url, {...options, agent: proxyAgent}, response => {
-      const statusCode = response.statusCode ?? 500;
-      if (!(statusCode >= 200 && statusCode < 300))
-        return reject(new Error(`Server answered with HTTP ${statusCode}`));
+      const statusCode = response.statusCode;
+      if (statusCode != null && statusCode >= 200 && statusCode < 300)
```
gao-sun, line 18: @aduh95 just curious why eqeq is allowed? usually js/ts projects force eqeqeq instead of eqeq to avoid implicit conversions.

aduh95: IIRC it’s allowed only for `null`, because it’s the only value for which it’s useful instead of being confusing.

PR #263, opened 2023-05-03
Updating DNS troubleshooting to mention ipv6

SpencerKaiser:
This issue is more widespread than I initially thought and some issues arise directly from DNS resolution of ipv6 domains, so I wanted to update the docs to call that out specifically

### DIFF
--- a/README.md
+++ b/README.md
@@ -224,7 +224,8 @@ network interaction.
 There are a wide variety of networking issues that can occur while running `corepack` commands. Things to check:
 
 - Make sure your network connection is active.
-- Make sure the host for your request can be resolved by your DNS; try using `curl [URL]` (ipv4) and `curl -6 [URL]` (ipv6) from your shell.
+- Make sure the host for your request can be resolved by your DNS; try using
+  `curl [URL]` (ipv4) and `curl -6 [URL]` (ipv6) from your shell.
 - Check your proxy settings (see [Environment Variables](#environment-variables)).
 
 ## Contributing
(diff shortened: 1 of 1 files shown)


markojak:
I want to add here that this is not a corner issue. There are over 32m spectrum customers and so far it seems all spectrum customers who have IPv6 enabled on their routers will be affected 

Evidently my router updated its firmware and then turned on IPv6 and then I had to stumble around for two days trying to figure out why Node18+ doesn't want to connect to the registry 

Even simple documentation like this can be helpful but I strongly suggest network level information on installs, ping and doctor

SpencerKaiser:
> Evidently my router updated its firmware and then turned on IPv6 and then I had to stumble around for two days trying to figure out why Node18+ doesn't want to connect to the registry

.... you don't have an Eero router, do you? 🤨 That is _**literally**_ what happened to me... overnight ipv6 got turned on, Spectrum as my provider, yarn + npm failures all over the place that were nearly impossible to debug.

README.md
```diff
@@ -224,7 +224,7 @@ network interaction.
 There are a wide variety of networking issues that can occur while running `corepack` commands. Things to check:
 
 - Make sure your network connection is active.
-- Make sure the host for your request can be resolved by your DNS; try using `curl [URL]` from your shell.
+- Make sure the host for your request can be resolved by your DNS; try using `curl [URL]` (ipv4) and `curl -6 [URL]` (ipv6) from your shell.
```
aduh95, line 227:
```suggestion
- Make sure the host for your request can be resolved by your DNS; try using
  `curl [URL]` (ipv4) and `curl -6 [URL]` (ipv6) from your shell.
```

aduh95 (approved):

PR #266, opened 2023-05-19
chore(deps): upgrade all npm dependencies

aduh95:


sources/corepackUtils.ts
```diff
@@ -85,6 +85,7 @@ export async function findInstalledVersion(installTarget: string, descriptor: De
     // If the dirname correspond to an in-range version and is not lower than
     // the previous best match (or if there is not yet a previous best match),
     // it's our new best match.
+    // @ts-expect-error TODO: decipher why this produces an error
     if (range.test(name) && maxSV?.compare(name) !== 1) {
```
aduh95, line 89: No idea why this code generates the following error:

```
error TS2339: Property 'compare' does not exist on type 'never'.

88     if (range.test(name) && maxSV?.compare(name) !== 1) {
```

merceyz:
```suggestion
    if (range.test(name) && maxSV && maxSV.compare(name) !== 1) {
```
This works so potentially a TypeScript bug?

tests/Disable.test.ts
```diff
@@ -1,5 +1,7 @@
+import {describe, beforeEach, it, expect}     from '@jest/globals';
 import {Filename, ppath, xfs, npath}          from '@yarnpkg/fslib';
-import {delimiter}                            from 'path';
+import {delimiter}                            from 'node:path';
+import process                                from 'node:process';
```
merceyz, line 4: These changes shouldn't be necessary, these are all global by default.

merceyz (changes requested):

tests/Disable.test.ts
aduh95: It's indeed not necessary, more like a personal preference.

aduh95: Fair enough, I thought at first it was necessary for this PR of the output of https://github.com/nodejs/corepack/actions/runs/5022431482/jobs/9005884303, turns out it was not and I left them because I liked them. My bad.

sources/corepackUtils.ts
aduh95:
```suggestion
    if (range.test(name) && (maxSV as semver.SemVer | undefined)?.compare(name) !== 1) {
```
This also don't trigger the error, so definitely a TS bug, I guess we just have to wait for them to fix it.

PR #270, opened 2023-05-26
feat: support ESM

merceyz:

- Corepack doesn't support package managers using ESM since it always loads them as CJS modules.
This isn't a problem at the moment but might be in the future.
- https://github.com/nodejs/corepack/pull/268 made Corepack preserve `process.exitCode` when the package manager throws synchronously which Node.js doesn't do by default.
- Corepack sends synchronously thrown errors by package managers to `stdout` instead of `stderr`.

**How did you fix it?**

Start package managers using `Module.runMain`.


arcanis:
Sounds good, but I think I'll merge it after the release, as it seems a little more dangerous than the two other changes

merceyz:
All the tests pass so this PR should be just as "safe" as what's currently in `main`.

PR #276, opened 2023-06-13
fix: disable `v8-compile-cache` when using `npm@>=9.7.0`

merceyz:
Node.js segfaults when using `npm@>=9.7.0` and `v8-compile-cache`
```console
$ docker run -it node:20.3.0-slim corepack npm@9.7.1 --version
[SIGSEGV]
```


jakebailey:
Aha, a repro! Thanks for finding this.

PR #283, opened 2023-07-13
chore: update dev dependency proxy-agent to 6.2.2

Trott:
This is to update the indirect vm2 dependency and resolve some dependabot alerts.

### DIFF
--- a/package.json
+++ b/package.json
@@ -41,7 +41,7 @@
     "eslint-plugin-arca": "^0.15.0",
     "jest": "^29.0.0",
     "nock": "^13.0.4",
-    "proxy-agent": "^6.0.0",
+    "proxy-agent": "^6.2.2",
     "semver": "^7.5.2",
     "supports-color": "^9.0.0",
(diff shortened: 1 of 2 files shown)


arcanis (approved):

merceyz:
This PR doesn't change the vm2 version, it's locked here https://github.com/nodejs/corepack/blob/6090fe7cbea7cebb1c6fee93d1ff417cb01da928/package.json#L99.

arcanis:
Ah right, we completely disabled vm2 to avoid this kind of problem. Still, the warnings from automation will be annoying.

I wonder if we could avoid it by using a link/portal rather than a patch, to completely prevent vm2 from being in the lockfile? 🤔

Trott:
Going to go ahead and land as it updates a few things and is hopefully harmless (and might quiet the automation). Feel free to revert or force push out if I'm messing anything up by doing that....

Trott:
Ugh, it didn't resolve a single warning... :-(

PR #284, opened 2023-07-14
chore: remove `vm2` from dev dependencies

aduh95:
vm2 is a transitive dependency that we do not use, and we patched it to be completely harmless. However, it seems the automation is not able to detect that, and is producing annoying warnings, so let's update it.

Refs: https://github.com/nodejs/corepack/pull/283


aduh95:
Hum it seems that 3.9.19 is also affected, so that won't fix the automated report 😒 https://www.npmjs.com/package/vm2 shows that the package is deprecated and won't be fixed, so let's remove it from our dependency tree instead.

aduh95:
It seems it's not possible to affect dependencies using patches, @arcanis is that right? I'd have sworn that was possible at some point.

arcanis:
Metadata are retrieved from the registry, so I think patch won't be able to change them; probably something we should fix in Yarn 🤔

In the meantime we can use a `resolutions` field to map `vm2` to a `portal:./path/to/vm2/mock`; I'm not in front of my computer to test but I think it should do the trick.

vm2/index.js
```diff
@@ -0,0 +1,5 @@
+"use strict";
+module.exports = {
+  VM: function VM() {},
+  VMScript: function VMScript() {},
```
aduh95, line 4:
```suggestion
  VM() {},
  VMScript() {},
```

merceyz (approved):

Issue #274, opened 2023-06-02
RFC: New command names (and alias the old ones)


arcanis:
I think the current command names are at odds with what most people would expect. What do you think about rewording them to be closer from what package managers' interfaces look like? The old commands would still work, but wouldn't be displayed in `--help` (or perhaps marked as deprecated in the documentation).

- `corepack prepare foo@1.2.3` -> `corepack install -g foo@1.2.3`
- `corepack prepare foo@1.2.3 --activate` -> `corepack use -g foo@1.2.3`

- `corepack hydrate path/to/archive.gz` -> `corepack install -g path/to/archive.gz`
- `corepack hydrate path/to/archive.gz --activate` -> `corepack use -g path/to/archive.gz`

- `corepack prepare -o ...` -> `corepack pack -g`

Along with new commands:

- `corepack use foo@1.2.3` -> Add the `packageManager` field to the `package.json` file
- `corepack remove -g foo` -> Removes all versions from `foo` from the cache
- `corepack up foo` -> Download the latest version from the registry and automatically use it
- `corepack info foo` -> List all versions currently in the cache (and the latest one from the remote)

ljharb:
`corepack remove foo` suggests uninstall to me; any reason not to nest all cache commands under `corepack cache`? eg `corepack cache remove foo`

arcanis:
I forgot to add the `-g` flag to `remove`, would that clarifies the intent? You may notice I also listed it as mandatory flag to `corepack install` to convey that it mutates the system, not the project. A command namespace like `cache` would be a possible alternative, but I'm worried most users would have troubles remembering it vs `-g`, which is very well known.

ljharb:
that still suggests a global uninstall to me, because there's more than one bucket of things that are mutated in the system - the global installs, the cache, config, and corepack itself, to name a few.

styfle:
> corepack use foo@1.2.3 -> Add the packageManager field to the package.json file

I would prefer `corepack add foo@1.2.3` to add/update the `packageManager` field in package.json

arcanis:
> I would prefer `corepack add foo@1.2.3` to add/update the `packageManager` field in package.json

I don't have a strong opinion on this; fwiw I used `corepack use` since adding a package manager would also remove one, which is different from package managers where adding a dependency doesn't mutate the others.

> that still suggests a global uninstall to me, because there's more than one bucket of things that are mutated in the system - the global installs, the cache, config, and corepack itself, to name a few.

[...]

PR #291, opened 2023-08-03
Refactoring of the CLI interface

arcanis:
This PR is the first step towards implementing the changes from https://github.com/nodejs/corepack/issues/274

- Implements `corepack install -g [--all] ...`
- Implements `corepack install`
- Implements `corepack use [pattern]`
- Implements `corepack up`
- Implements `corepack pack ...`

I added tests for `use` and `up` since those are new; `install` and `pack` should already covered by the existing tests, which I migrated from `prepare` / `hydrate` to the new commands.

The old commands are still there, but I removed their help messages so they don't appear in `corepack -h` anymore.


styfle:
> The old commands are still there

Thanks for keep those! That will help as we migrate support for multiple versions of corepack 👍

aduh95 (commented):
Can you please update the doc in the README?

sources/commands/Up.ts
```diff
@@ -0,0 +1,49 @@
...
+  static paths = [
+    [`up`],
+  ];
+
+  static usage = Command.Usage({
+    description: `Update the package manager used in the current project`,
+    details: `
+      Retrieve the latest available version for the current major release line of the package manager used in the local project, and update the project to use it.
+
+      Unlike \`corepack use\` this command doesn't accept/need a version range, as it will always select the latest available version from the same major line. Should you need to upgrade to a new major, use an explicit \`corepack use {name}@^{version}\` call.
```
aduh95, line 16: Would this work? If so, that's probably what we should document since it's the easiest to remember
```suggestion
      Unlike \`corepack use\` this command doesn't accept/need a version range, as it will always select the latest available version from the same major line. Should you need to upgrade to a new major, use an explicit \`corepack use {name}@latest\` call.
```

sources/commands/Use.ts
```diff
@@ -0,0 +1,38 @@
...
+  ];
+
+  static usage = Command.Usage({
+    description: `Define the package manager to use for the current project`,
+    details: `
+      Retrieve the latest release matching the provided descriptor, assign it to the project's package.json file, and automatically perform an install.
+    `,
+    examples: [[
+      `Configure the project to use the latest Yarn release`,
+      `corepack use 'yarn@*'`,
```
aduh95, line 17: I feel `latest` is more explicit than `*`, but no strong feeling
```suggestion
      `corepack use 'yarn@latest'`,
```

tests/Up.test.ts
```diff
@@ -0,0 +1,34 @@
...
+        packageManager: `yarn@2.1.0`,
+      });
+
+      await expect(runCli(cwd, [`up`])).resolves.toMatchObject({
+        exitCode: 0,
+        stderr: ``,
+      });
+
+      await expect(xfs.readJsonPromise(ppath.join(cwd, `package.json`))).resolves.toMatchObject({
+        packageManager: `yarn@2.4.3+8c1575156cfa42112242cc5cfbbd1049da9448ffcdb5c55ce996883610ea983f`,
```
aduh95, line 25: This lacks the algorithm name, shouldn't it be 
```suggestion
        packageManager: `yarn@2.4.3+sha224.8c1575156cfa42112242cc5cfbbd1049da9448ffcdb5c55ce996883610ea983f`,
```

sources/corepackUtils.ts
```diff
@@ -137,16 +148,26 @@ export async function installVersion(installTarget: string, locator: Locator, {s
 
   stream.pipe(sendTo);
 
-  const hash = build[0]
-    ? stream.pipe(createHash(build[0]))
-    : null;
-
+  const hash = stream.pipe(createHash(build[0] ?? `sha256`));
```
aduh95, line 151: I suggest we use `sha224` because it's shorter
```suggestion
  const hash = stream.pipe(createHash(build[0] ?? `sha224`));
```

sources/corepackUtils.ts
```diff
@@ -137,16 +148,26 @@ export async function installVersion(installTarget: string, locator: Locator, {s
...
 
-  const actualHash = hash?.digest(`hex`);
-  if (actualHash !== build[1])
+  const actualHash = hash.digest(`hex`);
+  if (build[1] && actualHash !== build[1])
     throw new Error(`Mismatch hashes. Expected ${build[1]}, got ${actualHash}`);
 
+  await fs.promises.writeFile(path.join(tmpFolder, `.corepack`), JSON.stringify({
+    locator,
+    hash: actualHash,
```
aduh95, line 160:
```suggestion
    hash: `sha224.${actualHash}`,
```

README.md
```diff
@@ -171,29 +170,52 @@ echo "function npx { corepack npx `$args }" >> $PROFILE
...
-| ------------ | ----------------------------------------- |
-| `--activate` | Also update the "Last Known Good" release |
+Retrieve the latest available version for the current major release line of
+the package manager used in the local project, and update the project to use
+it.
 
-This command will retrieve the given package manager from the specified archive
-and will install it within the Corepack cache, ready to be used without further
-network interaction.
+Unlike `corepack use` this command doesn't accept/need a version range, as it
```
aduh95, line 215:
```suggestion
Unlike `corepack use` this command doesn't take a package manager name nor a version range, as it
```

README.md
```diff
@@ -171,29 +170,52 @@ echo "function npx { corepack npx `$args }" >> $PROFILE
...
+will always select the latest available version from the same major line.
+Should you need to upgrade to a new major, use an explicit
+`corepack use {name}@^{version}` call.
```
aduh95, line 218:
```suggestion
`corepack use {name}@latest` call.
```

sources/corepackUtils.ts
arcanis: I'd have a small preference for sha256 because it's more well-known. I suspect most people seeing sha224 will wonder whether it's less secure and perhaps document themselves about it. sha256 seems less friction for users.

sources/commands/Up.ts
arcanis: Yarn doesn't have a `latest` tag available in Corepack, only `stable` and `canary` 🤔

We could hardcode a special case for that I guess.

sources/corepackUtils.ts
aduh95: It's easy to change if folks complain about it, but I suspect most folks won't pay attention to it – unless they're into software security, in which case I have the feeling they'd know what sha224 is. Until anyone complains `sha224` seems a better choice for me, no strong feelings though

README.md
aduh95:
```suggestion
`corepack use {name}@*` call.
```

sources/corepackUtils.ts
aduh95: Don't forget to merge this, or I guess the SHA-256 variant if we're not using SHA-224.
```suggestion
    hash: `sha256.${actualHash}`,
```

sources/corepackUtils.ts
styfle: > sha224 seems a better choice for me

Can you explain why?

aduh95: Because it's shorter, and more than strong enough for what that hash is used for.

arcanis: My intuition is that the extra characters won't change a lot the experience, but sha256 is more well-known. I kept it as-is for this iteration.

arcanis:
Should be good to merge.

Issue #95, opened 2022-03-02
Support for semver ranges?


jsumners:
I'd like to write a `package.json` like:

```json
{
  "packageJson": "pnpm@6"
}
```

Or maybe `pnpm@^6.32.2`, etc.

As it stands, adding a package manager to `package.json` requires a full specified version. This adds more maintenance overhead.

I think this is related to #93.

shellscape:
Agreed. pnpm v7 was released just yesterday and I'd hoped to use corepack moving forward instead of manually installing it for all of our workflows. This is probably done so along the lines of "pin your dependencies," but doesn't meet all needs.

styfle:
Would this need to perform a network request every time you do `pnpm run <script>` to check for the latest pnpm, since you can't rely the cached copy anymore?

aduh95:
It looks like this was originally possible, and was removed in https://github.com/nodejs/corepack/pull/18.

boneskull:
This seems kind of odd to restrict matching on exact version when one can provide any dist-tag instead (which can be moved around at will).  

Reverting the restriction would help my use case: use corepack to test or try something on a package manager where only the major version matters.  I have sinking feeling my use case is out of scope...

aduh95:
@boneskull The restriction only applies to the `package.json` ad hoc field, you can still do e.g. `corepack pnpm@7.x install`, does that cover your use case?

boneskull:
that wasn't working as of a couple hours ago. `corepack npm@7` unknown version or smth to that effect

boneskull:
wrapping `corepack`  to support this, I've pulled down the result of `npm show --json <npm|yarn|pnpm> versions`, then use [semver](https://npm.im/semver) to try match a range.

merceyz:
That seems like a bug `npm@7` isn't a range so it should expand to `npm@7.0.0`.

boneskull:
@aduh95 I wonder if the env var was the problem.  Seems to work now, thanks.

uiolee:
Totally agree. 

The exact version of the package manager prevents me from working on different projects. I don't think most projects need an exact package manager. corepack should introduce the semver strategy likes dependencies management.

mrgrain:
I always want to use the latest version of my package manager. They exception usually is on a new major release, when I first have to run a few validations.

With the current restrictions to only exact versions, I have to make a change every time I want to update the version of the package manger. Worst case, this is includes a PR and I need to bother a co-worker with a review.

[...]

PR #300, opened 2023-09-04
Don't require version range

### LINKED ISSUE #298: Corepack use should allow omitting version
The command `corepack use` requires specifying a version.

e.g. `corepack use 'yarn@*'`

This command should allow simply `corepack use yarn`.

rotu:
Fix #298

### DIFF
--- a/sources/commands/Base.ts
+++ b/sources/commands/Base.ts
@@ -14,7 +14,7 @@ export abstract class BaseCommand extends Command<Context> {
 
     const resolvedSpecs = all
       ? await this.context.engine.getDefaultDescriptors()
-      : patterns.map(pattern => specUtils.parseSpec(pattern, `CLI arguments`, {enforceExactVersion: false}));
+      : patterns.map(pattern => specUtils.parseSpec(pattern, `CLI arguments`));
 
     if (resolvedSpecs.length === 0) {
       const lookup = await specUtils.loadSpec(this.context.cwd);
--- a/sources/commands/InstallGlobal.ts
+++ b/sources/commands/InstallGlobal.ts
@@ -52,7 +52,7 @@ export class InstallGlobalCommand extends BaseCommand {
         if (arg.endsWith(`.tgz`)) {
           await this.installFromTarball(path.resolve(this.context.cwd, arg));
         } else {
-          await this.installFromDescriptor(specUtils.parseSpec(arg, `CLI arguments`, {enforceExactVersion: false}));
+          await this.installFromDescriptor(specUtils.parseSpec(arg, `CLI arguments`));
         }
       }
     } else {
--- a/sources/commands/deprecated/Prepare.ts
+++ b/sources/commands/deprecated/Prepare.ts
@@ -58,7 +58,7 @@ export class PrepareCommand extends Command<Context> {
 
     for (const request of specs) {
       const spec = typeof request === `string`
-        ? specUtils.parseSpec(request, `CLI arguments`, {enforceExactVersion: false})
+        ? specUtils.parseSpec(request, `CLI arguments`)
         : request;
 
       const resolved = await this.context.engine.resolveDescriptor(spec, {allowTags: true});
--- a/sources/specUtils.ts
+++ b/sources/specUtils.ts
@@ -1,27 +1,29 @@
 import {UsageError}                                     from 'clipanion';
 import fs                                               from 'fs';
 import path                                             from 'path';
-import semver                                           from 'semver';
 
 import {Descriptor, Locator, isSupportedPackageManager} from './types';
 
 const nodeModulesRegExp = /[\\/]node_modules[\\/](@[^\\/]*[\\/])?([^@\\/][^\\/]*)$/;
 
-export function parseSpec(raw: unknown, source: string, {enforceExactVersion = true} = {}): Descriptor {
+export function parseSpec(raw: unknown, source: string): Descriptor {
   if (typeof raw !== `string`)
     throw new UsageError(`Invalid package manager specification in ${source}; expected a string`);
 
-  const match = raw.match(/^(?!_)(.+)@(.+)$/);
-  if (match === null || (enforceExactVersion && !semver.valid(match[2])))
-    throw new UsageError(`Invalid package manager specification in ${source}; expected a semver version${enforceExactVersion ? `` : `, range, or tag`}`);
+  const match = /^(@?[^@]+)(?:@(.+))?$/.exec(raw);
+  const name = match?.[1];
+  const range = match?.[2] ?? `*`;
 
-  if (!isSupportedPackageManager(match[1]))
+  if (!name) {
+    throw new UsageError(
+      `Invalid package manager specification in ${source}. Could not determine package manager name`,
+    );
+  }
+
+  if (!isSupportedPackageManager(name))
     throw new UsageError(`Unsupported package manager specification (${match})`);
 
-  return {
-    name: match[1],
-    range: match[2],
-  };
+  return {name, range};
 }
 
 /**


merceyz (changes requested):
This allows omitting the version from the `package.json` `packageManager` field as well which it shouldn't do.

arcanis:
And for `corepack install -g` (and this command only), `yarn` should alias to `yarn@^1`, not `yarn@*`.

rotu:
> This allows omitting the version from the `package.json` `packageManager` field as well which it shouldn't do.

Hmm... I thought that was enforced in `resolveDescriptor`, which has the `allowTags` option. Under what circumstances (if any) would we allow a version range but *not* a tag?

rotu:
> And for `corepack install -g` (and this command only), `yarn` should alias to `yarn@^1`, not `yarn@*`.

@arcanis I left that out of scope on purpose. I can't figure out the intended behavior so I punted.

At the very least it seems like `corepack use yarn` should be synonymous with `corepack use yarn@*` and either `yarn@stable` or `yarn@latest`.

But here's what it is right now:

* `corepack install -g yarn@stable` -> yarn 3.6.3
* `corepack install -g 'yarn@*'` -> yarn 4.0.0-rc.50
* `corepack install -g 'yarn@latest'` -> Usage Error: Tag not found (latest)

* `npm exec --package yarn -c 'yarn --version'` -> 1.22.19
* `npm exec --package 'yarn@*' -c 'yarn --version'` -> 1.22.19
* `yarn dlx --package 'yarn' yarn --version` -> 1.22.19
* `yarn dlx --package 'yarn@*' yarn --version` -> 2.4.3

---

Heck, even https://www.npmjs.com/package/yarn and https://yarnpkg.com/package?q=yarn&name=yarn don't acknowledge any version of yarn exists past 2.4.3, so how should a user know what a valid version specifier is?

Yarn is a wonderful project but the critical user story of "how should I install it, and what version should I use?" is a dumpster fire. I am of the opinion that the yarn cli needs to  provide its own backwards compatibility affordances, not require other tools to implement special rules.

arcanis:
Yeah, I think it's fine to keep it as a follow-up. I have another idea to make this process a little less magic, I'll handle that.

> Heck, even https://www.npmjs.com/package/yarn and https://yarnpkg.com/package?q=yarn&name=yarn don't acknowledge any version of yarn exists past 2.4.3, so how should a user know what a valid version specifier is?

Because you're looking at the npm package, and [Yarn hasn't been distributed on npm for the past couple of years](https://yarnpkg.com/getting-started/qa#why-is-the-yarn-package-on-npm-still-on-1x). So yeah, the version you're seeing there is outdated. The proper list of versions is here: https://repo.yarnpkg.com/tags

rotu:
> Because you're looking at the npm package, and [Yarn hasn't been distributed on npm for the past couple of years](https://yarnpkg.com/getting-started/qa#why-is-the-yarn-package-on-npm-still-on-1x). So yeah, the version you're seeing there is outdated. The proper list of versions is here: https://repo.yarnpkg.com/tags

That's awesome documentation and makes a lot of sense!

The [npm page](https://www.npmjs.com/package/yarn) and [yarn page](https://yarnpkg.com/package?name=yarn) are the de facto storefront for a javascript package, and even if they aren't the place releases happen, they need guidance in the readmes (even if those instructions start with "Don't install this version and don't install with npm" and a link to the recommendations for how to get started with yarn).

The npm page has also been confusing because the "berry" dist-tag implies that "berry" is the name for major version 2, but the active development repo is also called "berry" and has releases for major versions 2, 3, and 4.

I have definitely been turned off (perhaps unfairly) from trying yarn in the past by these sorts of issues!

rotu:
> This allows omitting the version from the `package.json` `packageManager` field as well which it shouldn't do.

@merceyz Why not?

Please see my latest commit, which is how I feel it should work, somewhat like `npm install --save-exact`.

aduh95:
@merceyz @arcanis Can we move forward with this?

merceyz:
> @merceyz Why not?

Reproducibility, if a project is installable and works today it should work in a year, if a range is used that might not be the case anymore.

> @merceyz @arcanis Can we move forward with this?

@aduh95 I don't think this should land in its current state and it needs some tests to ensure it works as expected.

rotu:
> > @merceyz Why not?
> 
> Reproducibility, if a project is installable and works today it should work in a year, if a range is used that might not be the case anymore.

Note that `corepack use npm@stable` is accepted and has the same issue. This locks down the version in the same way.

merceyz:
That command would set the `package.json#packageManager` field to a specific version of npm, not a range, which is fine.

rotu:
> That command would set the `package.json#packageManager` field to a specific version of npm, not a range.

That's my understanding too - I feel like I don't understand your point. I'm picturing the user specifying a value loosely with `corepack use`. Then `corepack install` will set up the project locked to that particular package manager.

> if a range is used that might not be the case anymore.

What's the use case you're envisioning here which, after this PR, would create a non-reproducible build?

rotu:
@merceyz updated the tests. I think this clarifies how I expect this to work, basically like `npm install --save-exact`, where the user can be fuzzy but the result will be specific.

arcanis (changes requested):
> Why shouldn't we trust them? It might be more important to the user to have bug fixes than to have reproducibility, especially for nascent package managers like bun.

The current model is to favour good practices, in particular reproducibility. It's a major point of Corepack, otherwise you can just npm install whatever package manager you want and that's it, no need for Corepack.

I agree with @merceyz on this - adding support for loose specifiers is fine on the CLI commands (because they only matter at the time the commands are run), but allowing them in the `packageManager` field is a different story that should be evaluated in its own PR (and I'll push back on it for the reasons described above).

rotu:
> > Why shouldn't we trust them? It might be more important to the user to have bug fixes than to have reproducibility, especially for nascent package managers like bun.
> 
> The current model is to favour good practices, in particular reproducibility. It's a major point of Corepack, otherwise you can just npm install whatever package manager you want and that's it, no need for Corepack.

Good practice is, when reproducibility is key, to track the full version with hash, which is exactly what `corepack use` does. If the package manager actually follows the semantic versioning spec, then good practice is to stick with compatible releases, including tracking prerelease lines as they mature.

Even without this restriction, corepack is necessary because:

1. Typically `node_modules/.bin` is not in `$PATH` so even if you install your preferred package manager for a project, it's unwieldy to call.
2. The `yarn` package manager does not make itself available in a package repository. That is, `npm install yarn@latest` and `npm install yarn@berry` install versions which are very out of date.

> I agree with @merceyz on this - adding support for loose specifiers is fine on the CLI commands (because they only matter at the time the commands are run), but allowing them in the `packageManager` field is a different story that should be evaluated in its own PR (and I'll push back on it for the reasons described above).

That's going to wind up with those CLI commands in build scripts and commit hooks. The whole idea behind a human-editable `package.json` is to track the developer's *intent*, not enforce reproducibility (that's what the lockfile is for). #95 has 28 👍 versus 2 👎, which seems to clearly demonstrate that we should accept loose versions here.

@arcanis and @merceyz, would you be happy with instead [reifying the version on `corepack install`](https://github.com/nodejs/corepack/pull/300/commits/6cf4ea9bbe0cf0d2f2ed04c3b00007ba1b1fe38c) as I previously had?

rotu:
> > @arcanis and @merceyz, would you be happy with instead [reifying the version on `corepack install`](https://github.com/nodejs/corepack/pull/300/commits/6cf4ea9bbe0cf0d2f2ed04c3b00007ba1b1fe38c) as I previously had?
> 
> Not sure what you mean by that - as far as I can tell from the tests you link, this still lets you write a range in the `packageManager` field, which is what I don't think we should merge as part of this PR.

Yes. In that revision of the PR, this lets you write a range in the `packageManager` field and then, the first time you `corepack install`, it rewrites the `packageManager` field to the specific version. This is a much more comfortable dev experience than crashing out with a "Usage Error" (even on `corepack use` and `corepack up`!).

> Essentially, I believe this PR is conflating two different things (ranges in the CLI, and ranges in the project definition), and I'm against using the merits of the first request (which has consensus) to land the second (which hasn't).

Yes, in fixing the first, I realized how arbitrary and unwieldy the second one is and what it would take to fix. The feedback on #18 and #95 seems clear consensus that having something `package.json` which *looks like* a package specifier but doesn't work like one is bad DX, and that users should be able to edit `package.json` by hand.

So @arcanis and @merceyz, what do you think the developer's EXPECTATION is if they have written an imprecise version (or none at all) into package.json?

Issue #303, opened 2023-09-04
Tests assume `COREPACK_ENABLE_STRICT` unset


rotu:
In my shell profile, I export `COREPACK_ENABLE_STRICT=0`. This causes the test to fail, since the "should refuse to run a different package manager within a configured project" test assumes this environment variable is unset.

https://github.com/nodejs/corepack/blob/2bf76c9e29a59dee21e96cf6250a8cede59922e5/tests/main.test.ts#L262C28-L262C28

styfle:
This should fix it https://github.com/nodejs/corepack/pull/304

PR #304, opened 2023-09-04
chore: update test to handle case when env var is set

styfle:
- Fixes https://github.com/nodejs/corepack/issues/303

### DIFF
--- a/tests/main.test.ts
+++ b/tests/main.test.ts
@@ -260,7 +260,6 @@ it(`should allow to call "corepack install" without arguments within a configure
 });
 
 it(`should refuse to run a different package manager within a configured project`, async () => {
-
   await xfs.mktempPromise(async cwd => {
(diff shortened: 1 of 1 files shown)


tests/main.test.ts
```diff
@@ -260,6 +260,11 @@ it(`should allow to call "corepack install" without arguments within a configure
 });
 
 it(`should refuse to run a different package manager within a configured project`, async () => {
+  // To ensure we test the default behavior, we must remove these env vars
+  // in case the local machine already set these values.
+  delete process.env.COREPACK_ENABLE_STRICT;
+  delete process.env.FORCE_COLOR;
```
arcanis, line 266: We should move it to [`setupTests.js`](https://github.com/nodejs/corepack/blob/main/tests/setupTests.js), to ensure a consistent environment for all tests.

styfle: Good idea! I pushed an update 👍

PR #307, opened 2023-09-11
[WIP] Add support for bun

### LINKED ISSUE #295: Bun support
Hi there,

I have no relation to the bun team at all (and, am a bit daft on web stuff). I'm looking to change to bun from pnpm for performance reasons, but there's no way to do this through the corepack system today (which has been very kind to me, for what it's worth).

It would be very nice if I could use the corepack mechanism to change to bun, similar to how (p)npm are implemented today from yarn.

https://github.com/oven-sh/bun
https://bun.sh/

It looks like they have a 1.0 mark coming on September 7th 2023.

Ethan-Arrowood:
Work in progress - adding support for Bun. I filled out the json as best I see fit.

Investigating tests next

Closes #295


Ethan-Arrowood:
I don't know why CI is failling at basic install step. Works fine locally for me

uncenter:
> I don't know why CI is failling at basic install step. Works fine locally for me

Looks like those tests are trying to install yarn?

image

merceyz:
Node.js v20.6 [changed the ESM loader a bit](https://github.com/nodejs/node/pull/47999) so the version of Yarn used by Corepack needs to be updated to a version containing https://github.com/yarnpkg/berry/pull/5677.

Should be a separate PR but you can test with:
```sh
yarn set version canary
yarn
```

Jarred-Sumner:
Would this be a symlink that does not execute JS in node? I think we should have a hard requirement that if Bun is to be included in Corepack, it must not incur the startup cost penalty of Node before executing `bun` (outside of the initial run to lazily install it)

arcanis:
All calls to all commands have to go through Corepack, since it has to read the packageManager field to know the version of the package manager to use. It doesn't "cache" this information.

aduh95:
@Ethan-Arrowood is it something you're still working on? Is this block on something?

Ethan-Arrowood:
Hey, no I'm not actively working on this anymore. It looks like we won't get sign off from Bun unless core pack is modified (https://github.com/nodejs/corepack/pull/307#issuecomment-1718747222). I'm happy to help move this along if there is a clear path forward though.

aduh95:
FWIW Corepack focuses on improving DX rather than performance, so IMO it wouldn't make sense to have a hard requirement as suggested in https://github.com/nodejs/corepack/pull/307#issuecomment-1718747222. Furthermore I don't think we need Bun permission (although if they feel strongly that we should not do that, it would be rude to not listen, but IIUC it's not what https://github.com/nodejs/corepack/pull/307#issuecomment-1718747222 is saying, please correct me if I'm wrong), so if the community wants Bun in Corepack, that's all that's needed (and passing CI of course). Contributions to improve Corepack performance are of course welcome, but it's not blocking as far as Corepack is concerned.

The path forward is to mark the PR as ready for reviews, and since the tests are already passing, get at least one approval and no objections.

Ethan-Arrowood:
Okay, I'll mark as ready for review now. And let folks review/approve/reject as they see fit.

Ethan-Arrowood:
I think this pr may need some tests of its own so if any core pack maintainers could help me with that. It would be appreciated

merceyz (changes requested):
This needs tests to ensure it works.

altechzilla:
Hey friends, any update on corepack support for bun? Really looking forward to this. Thanks!

aduh95:
> @merceyz can you help me with the tests? I don't understand what to add where

@Ethan-Arrowood you can use https://github.com/nodejs/corepack/pull/333 for inspiration, I think we just need tests that validates the download succeeds and bun is able to run.

Ethan-Arrowood:
Sorry for the delay. I've added more tests to match that other PR.

aduh95:
Bun package on the npm registry is relying on optional dependencies (https://www.npmjs.com/package/bun?activeTab=code). Because Corepack is not a package manager, it has no support for dependencies, so it doesn't work. Worth noting the `bun` binary tries to spawn a subprocess that expects `npm` to be on the $PATH, which seems problematic in the context of Corepack – I would consider it a bug if Corepack needed `npm` to be installed on the PATH to operate.

Support for Bun is still possible to implement into Corepack, but probably not from the npm package.

Note you also need to update the following list: https://github.com/nodejs/corepack/blob/12f1c31a0decb867f91f31ffb0cadf34ce9f0673/sources/types.ts#L4-L8

Ethan-Arrowood:
I'm not sure I understand, so we need to find a different url for installing bun? Would their install script suffice `https://bun.sh/install` ? 

I'm also going to re-emphasize that I really don't need to be driving this anymore. If this is more complex, I'd rather the Bun folks finish the swing here.

MinskLeo:
Hi guys! What's the latest status of adding bun support? Feels like it's kinda hard to add and it's blocked with smth? What needs to be done in order to add bun support into Corepack at this moment?

PR #308, opened 2023-09-11
chore: update Yarn

merceyz:
Node.js v20.6 [changed the ESM loader a bit](https://github.com/nodejs/node/pull/47999) so the version of Yarn used by Corepack needs to be updated to a version containing https://github.com/yarnpkg/berry/pull/5677.

Fixes the install failing on Node.js v20.

Ref https://github.com/nodejs/corepack/pull/307#issuecomment-1714464178

### DIFF
--- a/package.json
+++ b/package.json
@@ -16,7 +16,7 @@
     "./package.json": "./package.json"
   },
   "license": "MIT",
-  "packageManager": "yarn@4.0.0-rc.44+sha224.6526204ca38ed0105e81ba52d83dc0c7b8ee63600a13dc332914fde0",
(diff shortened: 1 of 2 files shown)


merceyz:
The release candidate of Yarn v4 that contains the fix has dropped support for Node.js v16 so we'll either need to do the same, downgrade to Yarn v3, or backport the fix to a v4 release candidate that still supported v16.

arcanis (approved):
Node.js v16 is officially EOL starting from today(!), so we probably can remove it from the CI

Ethan-Arrowood (approved):

PR #309, opened 2023-09-11
chore: update supported Node.js versions

aduh95:
BREAKING CHANGE: remove support for Node.js 16.x

### DIFF
--- a/tsconfig.json
+++ b/tsconfig.json
@@ -12,7 +12,7 @@
     "resolveJsonModule": true,
     "skipLibCheck": true,
     "strict": true,
-    "target": "ES2023"
+    "target": "ES2022"
   },
   "ts-node": {
     "transpileOnly": true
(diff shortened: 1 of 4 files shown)


tsconfig.json
```diff
@@ -7,12 +7,12 @@
...
     "noEmit": true,
     "forceConsistentCasingInFileNames": true,
-    "lib": ["ES2021"],
+    "lib": ["ES2023"],
     "module": "commonjs",
     "resolveJsonModule": true,
     "skipLibCheck": true,
     "strict": true,
-    "target": "ES2021"
+    "target": "ES2023"
```
styfle, line 15: I don't think `target` works with `ES2023` yet. You can try `ES2022`

```suggestion
    "target": "ES2022"
```

Or you can use `ESNext` so that bump `typescript` will also bump the target.

davyus321311:
> BREAKING CHANGE: remove support for Node.js 16.x

For the god give me Back my freedom for ALL my family

PR #312, opened 2023-10-08
test: check that a semver version is required

merceyz:
Adds a test to ensure that not specifying a semver version in `package.json#packageManager` isn't allowed.
Ref https://github.com/nodejs/corepack/pull/300#pullrequestreview-1608951980.


rotu:
How does this work with package managers that use non-semver versioning or for package specifiers like `npm@github:npm/cli#c12ea07` or `yarn@https://registry.npmjs.org/yarn/-/yarn-2.4.3.tgz`?

aduh95 (approved):

merceyz:
It doesn't, neither of those are supported.
Note that this PR doesn't change any behaviour it merely tests existing behaviour.

Issue #329, opened 2023-11-14
`corepack install` fails if there is no `$HOME` directory


SimenB:
```Dockerfile
FROM ruby:2.7.8-alpine3.16

RUN apk update && \
  apk add --no-cache nodejs npm shadow && \
  rm -rf /var/cache/apk/*

RUN npm install -g corepack@0.20.x

RUN mkdir /hello

RUN useradd --uid 1000 user

RUN chown -R 1000 /hello

USER 1000

RUN corepack enable

WORKDIR /hello

RUN echo '{"packageManager": "yarn@1.22.20+sha256.f283c286f2731dc029203ff77bca71ce1a0db13174fe8cf61c9e9a917215ff02"}' > package.json

RUN corepack install
```

Gives the following error:

```
0.242 Adding yarn@1.22.20+sha256.f283c286f2731dc029203ff77bca71ce1a0db13174fe8cf61c9e9a917215ff02 to the cache...
0.253 Internal Error: EACCES: permission denied, mkdir '/home/user/.cache/node/corepack'
0.253     at mkdirSync (node:fs:1382:3)
0.253     at getTemporaryFolder (/usr/local/lib/node_modules/corepack/dist/lib/corepack.cjs:38948:27)
0.253     at installVersion (/usr/local/lib/node_modules/corepack/dist/lib/corepack.cjs:39140:21)
0.253     at async Engine.ensurePackageManager (/usr/local/lib/node_modules/corepack/dist/lib/corepack.cjs:39350:32)
0.253     at async InstallLocalCommand.execute (/usr/local/lib/node_modules/corepack/dist/lib/corepack.cjs:39833:5)
0.253     at async InstallLocalCommand.validateAndExecute (/usr/local/lib/node_modules/corepack/dist/lib/corepack.cjs:37292:22)
0.253     at async Cli.run (/usr/local/lib/node_modules/corepack/dist/lib/corepack.cjs:38267:18)
0.253     at async Object.runMain (/usr/local/lib/node_modules/corepack/dist/lib/corepack.cjs:40228:12)
------
Dockerfile:23
--------------------
  21 |     RUN echo '{"packageManager": "yarn@1.22.20+sha256.f283c286f2731dc029203ff77bca71ce1a0db13174fe8cf61c9e9a917215ff02"}' > package.json
  22 |     
  23 | >>> RUN corepack install
  24 |     
--------------------
[...]

PR #343, opened 2023-12-28
Corepack "EACCES": Clearer Errors

Medhansh404:
#329 
Targeted Error Handling: Code distinguishes between EEXIST (not relevant) and EACCES (triggers the specific error message), while other errors are still thrown for debugging.
Implementation: Changes are made in the getTemporaryFolder function within node_modules/corepack/dist/lib/corepack.cjs.
Testing: not done.


sources/folderUtils.ts
```diff
@@ -52,9 +52,11 @@ export function getTemporaryFolder(target: string = tmpdir()) {
     } catch (error) {
       if ((error as NodeError).code === `EEXIST`) {
         continue;
+      } else if ((error as NodeError).code === `EACCES`) {
+        throw new Error(`Failed to create cache directory. Please ensure the user has write access to the target directory (${target}). If the user's home directory does not exist, create it first.`);
```
aduh95, line 56:
```suggestion
        throw new Error(`Failed to create cache directory. Please ensure the user has write access to the target directory (${target}). If the user's home directory does not exist, create it first.`, {cause: error});
```

arcanis: It'd be even better to throw a `UsageError`. The stacktrace is useless here.

Medhansh404: Is there anything i can do here i mean is there a custom error class `UsageError` required such as this 
`class UsageError extends Error {
  target: string;
  constructor(message: string, target: string, cause?: unknown) {
    super(message);
    this.target = target;
    this.cause = cause;
    Object.setPrototypeOf(this, UsageError.prototype);
}
}` (just a refernce), I can work on that if that is needed at all :). I would appreciate a feedback!

aduh95: You can import it:
https://github.com/nodejs/corepack/blob/fe3e5cd86c45db0d87c7fdea87d57d59b0bdcb78/sources/Engine.ts#L1

aduh95: It's likely that the original error stack would help in debugging though, I don't know if `UsageError` is really helpful here.

arcanis: Since this error is only present once in the source code, we will know what threw it if necessary. And for regular users, the stacktrace is typically unactionable information that makes it look like the program had an internal bug (vs a known limitation like here).

aduh95: Well you're making assumptions that you know what caused it, but it might be some other edge case that we didn't consider, and not having a stack trace is only going to make debugging harder in this case. Corepack is meant to be used by devs, and I would expect that most users would be more comfortable with a stack trace than without one.
That being said, I don't feel strongly, so if @Medhansh404 you can implement one or the other suggestion, and remove the unrelated spacing changes, we can land and make a new release.

Medhansh404: well I meant it in more of a user googling the error or checking on stack overflow, stacktrace would have made sense for similar searches in that context, but here the message is pretty self-explanatory so it works perfectly too whether the regukar users or new ones, I'll also take a look at the `UsageError` meanwhile.

arcanis: I don't think this will give us valuable information. The `fs` error messages are static and tied to the error code, so the message will only ever be a known one (the `EACCES` one, since we gate the exception on it). Knowing the code path isn't relevant either, since `EACCES` happens based on the state of the disk, not the code followed to attempt the syscall.

From my experience, anything that makes user errors easier to identify and understand is valuable - even if the audience are developers, not all developers have the same debug proficiency, and removing noise (when truly unhelpful like here, imo) contributes to clear up potential confusions.

aduh95: It might be the case today, but may change later in future versions of Node.js. In reality, this will hardly ever matter, since this is an edge case anyway, so both sound good to me.

Medhansh404: yup !! this sounds good to me I have made the changes; the discussion makes sense. Also, I went with the `UsageError` suggestion and it is up for reviewing @aduh95

sources/folderUtils.ts
```diff
@@ -52,9 +53,11 @@ export function getTemporaryFolder(target: string = tmpdir()) {
...
         continue;
+      } else if ((error as NodeError).code === `EACCES`) {
+        throw new UsageError(`Failed to create cache directory. Please ensure the user has write access to the target directory (${target}). If the user's home directory does not exist, create it first.`, {cause: error});
       } else {
         throw error;
       }
     }
   }
-}
+}
```
aduh95, line 63:
```suggestion
}

```

sources/folderUtils.ts
```diff
@@ -52,6 +53,8 @@ export function getTemporaryFolder(target: string = tmpdir()) {
     } catch (error) {
       if ((error as NodeError).code === `EEXIST`) {
         continue;
+      } else if ((error as NodeError).code === `EACCES`) {
+        throw new UsageError(`Failed to create cache directory. Please ensure the user has write access to the target directory (${target}). If the user's home directory does not exist, create it first.`, {cause: error});
```
aduh95, line 57: @arcanis it would maybe make sense to add support for this in Clipanion, no?

aduh95:
```suggestion
        // @ts-expect-error UsageError expects one argument
        throw new UsageError(`Failed to create cache directory. Please ensure the user has write access to the target directory (${target}). If the user's home directory does not exist, create it first.`, {cause: error});
```

Medhansh404: For now should we remove the `cause: error`?

aduh95:
There are still lint errors to address.

aduh95 (approved):

Issue #347, opened 2024-01-04
`corepack use` fails with ENOENT: no such file or directory, open '/app/package.json'


styfle:
When a package.json file is missing, `corepack use` fails with ENOENT.

I think it would be great if corepack could create the package.json file when its missing.

It would avoid the need to call `npm init -y`, `yarn init -y`, or `pnpm init` in some cases.

aduh95:
What do you think should be the behavior? Should Corepack call `<pkg-manager> init` in such case? Or just create a `package.json` with only the `"packageManager"` field?

styfle:
The second option - create package.json directly

styfle:
@aduh95 Thanks for the quick fix! 🎉

PR #350, opened 2024-01-06
fix(use): create `package.json` when calling `corepack use` on empty dir

aduh95:
Fixes: https://github.com/nodejs/corepack/issues/347


sources/specUtils.ts
```diff
@@ -96,10 +100,15 @@ export async function loadSpec(initialCwd: string): Promise<LoadSpecResult> {
...
+    try {
+      file = await fs.promises.open(manifestPath, `r`);
+    } catch (err) {
+      if ((err as NodeError)?.code === `ENOENT`) continue;
+      throw err;
+    }
 
-    const content = await fs.promises.readFile(manifestPath, `utf8`);
+    const content = await file.readFile(`utf8`);
+    await file.close();
```
arcanis, line 111: Can't we just use `readFile` and assign `{}` to `data` wen catching `ENOENT`?

aduh95: Sure, the only anti-pattern is to rely on `existsSync`, as its result might be outdated by the time the `readFile` calls is processed.

merceyz (commented):
Could you add a test for what should happen when running something like
```sh
corepack use foo@1
mkdir src
cd src
corepack use foo@2
```

sources/commands/Use.ts
```diff
@@ -16,7 +16,7 @@ export class UseCommand extends BaseCommand {
     `,
     examples: [[
       `Configure the project to use the latest Yarn release`,
-      `corepack use 'yarn@*'`,
+      `corepack use 'yarn'`,
```
arcanis, line 19:
```suggestion
      `corepack use yarn`,
```

sources/specUtils.ts
```diff
@@ -96,10 +99,12 @@ export async function loadSpec(initialCwd: string): Promise<LoadSpecResult> {
       continue;
 
     const manifestPath = path.join(currCwd, `package.json`);
-    if (!fs.existsSync(manifestPath))
-      continue;
-
-    const content = await fs.promises.readFile(manifestPath, `utf8`);
+    try {
+      content = await fs.promises.readFile(manifestPath, `utf8`);
+    } catch (err) {
```
arcanis, line 104: nit: The `content` declaration should be right above the `try/catch`, so it's clearer where it's assigned
```suggestion
    let content: string;
    try {
      content = await fs.promises.readFile(manifestPath, `utf8`);
    } catch (err) {
```

arcanis (approved):
Couple of nits, but LGTM overall

sources/specUtils.ts
```diff
@@ -88,6 +89,8 @@ export async function loadSpec(initialCwd: string): Promise<LoadSpecResult> {
     manifestPath: string;
   } | null = null;
 
+  let content: string;
+
```
aduh95, line 93:
```suggestion
```

merceyz (approved):
LGTM.

NIT: This seems more like a new feature (`feat:`) than a bugfix (`fix:`).

merceyz:
Right, in that case I agree, it's a bugfix 👍

Issue #104, opened 2022-05-03
Project Status (experimental to stable?)


styfle:
I can't find any details on the current project status such as anything that needs to be solved before going stable.

1. Are there any outstanding blockers?
2. Are there known breaking changes coming up?
3. Do we need to get npm working to be considered stable?
4. Will a stable release remove the need to run `corepack enable` or will that always be a requirement?

andersk:
I’d hope for security that corepack starts checking integrity hashes before it’s enabled by default.

- #37

trivikr:
Visiting this GitHub issue from [twitter thread](https://twitter.com/styfle/status/1526605695436587008).

I'd asked on [Yarn Discord](https://discord.com/channels/226791405589233664/226791405589233664/936365371571703898) in Jan'22 if we should propose to make it stable in Node.js 18. There was no follow-up though.

I'd shared some useful links:
* Corepack was backported to `v14.19.0` in https://github.com/nodejs/node/pull/41696
* https://github.com/nodejs/node/issues/14532
* https://github.com/nodejs/node/issues/19453

@arcanis have been the champion and lead author for corepack, along with other maintainers of yarn.
They're currently busy with the launch of yarn v4 in https://github.com/yarnpkg/berry/issues/3591

I think we should revive this GitHub issue after yarn v4 release.

arcanis:
> Are there known breaking changes coming up?

Mostly just #93 - many Node contributors expressed interest in seeing Corepack have a slightly more dynamic behavior, so that it stay up-to-date with minor/patch versions from the package managers (look at the thread for details, but in short it's currently painful for Node to have to make new releases each time npm has a security update).

It makes dealing with #37 a little harder though. Perhaps package managers releases could be signed instead, although it requires infra and can be be difficult to do for package managers that aren't distributed as a single JS bundle (I believe only Yarn is).

> Are there any outstanding blockers?

Apart from the one described above, I'd personally like to have feedback from one or more cloud provider confirming that Corepack has all the tools they need to build a proper integration that doesn't break their customers' setup. That's something Vercel can certainly help with!

> Do we need to get npm working to be considered stable?

Corepack has value for Yarn and pnpm users, and it'd be too bad if we were blocked going forward by something we have no control over, so in my opinion I'd tend to say it's not needed, although it would be nice to have.

Note that technically Corepack already supports npm, it's just not part of the default distribution until they decide to enable it. So when that happens it will be fairly fast and won't require much engineering work.

[...]

PR #351, opened 2024-01-08
feat!: remove `--all` flag

aduh95:
Force users to opt-in into the package managers they want to support. As more and more package managers are added to Corepack, it's more and more unlikely that our users will be interested in _all_ the package managers Corepack supports, and it is unreasonable to expect Corepack maintainers would be able to perform security audits.


arcanis (changes requested):
I find the increased friction in user experience a little problematic. Running `corepack enable` and things Just Work is easy. having to run the command multiple times isn't, especially if the only message people get is "command not found". Can we find alternatives?

aduh95:
Yes unfortunately it is likely some script out there relies on it. We could deprecate `corepack enable`, freeze it to always mean `corepack enable pnpm@latest yarn@1.x` (i.e don't add more package manager to it), and possibly make it emit a warning.

I expect the `--all` flag is good to go, no?

arcanis:
> I expect the --all flag is good to go, no?

The `--all` flag was intended to help build base offline base images for cloud providers, which frequently want all tools to be included in their images. I don't have a strong opinion as to whether this is super useful or not, but I'd tend to refrain from removing it, to avoid unneeded friction.

aduh95:
Maybe @styfle would have an opinion there? I would expect Cloud provider to be "picky" about which package managers they enable by default, as more and more package managers are added to Corepack, I would expect they want to activate only a subset.

styfle:
> I would expect they want to activate only a subset.

At Vercel, we only enable the package manager found in package.json.

I didn't know the `--all` flag existed until now.

I think we need to look at more long term and see how this would impact the scenario when `corepack enable` is no longer needed, presumably when it goes stable https://github.com/nodejs/corepack/issues/104

arcanis (approved):
Not 100% convinced, but I don't have a very strong opinion either way - it can be added back later on if really needed.

nickmccurdy:
How do we run upgrade package managers like we could before this change?

`corepack prepare --activate pnpm` works, but `corepack prepare --activate yarn` errors with:

```
Internal Error: Error when performing the request to https://registry.npmjs.org/yarn; for troubleshooting help, see https://github.com/nodejs/corepack#troubleshooting
    at fetch (/usr/local/Cellar/corepack/0.25.2/libexec/lib/node_modules/corepack/dist/lib/corepack.cjs:22882:11)
    at process.processTicksAndRejections (node:internal/process/task_queues:95:5)
    at async fetchAsJson (/usr/local/Cellar/corepack/0.25.2/libexec/lib/node_modules/corepack/dist/lib/corepack.cjs:22896:20)
    at async fetchAvailableVersions (/usr/local/Cellar/corepack/0.25.2/libexec/lib/node_modules/corepack/dist/lib/corepack.cjs:22960:20)
    at async fetchAvailableVersions2 (/usr/local/Cellar/corepack/0.25.2/libexec/lib/node_modules/corepack/dist/lib/corepack.cjs:22999:14)
    at async /usr/local/Cellar/corepack/0.25.2/libexec/lib/node_modules/corepack/dist/lib/corepack.cjs:23428:25
    at async Promise.all (index 0)
    at async Engine.resolveDescriptor (/usr/local/Cellar/corepack/0.25.2/libexec/lib/node_modules/corepack/dist/lib/corepack.cjs:23425:22)
    at async PrepareCommand.execute (/usr/local/Cellar/corepack/0.25.2/libexec/lib/node_modules/corepack/dist/lib/corepack.cjs:24135:24)
    at async PrepareCommand.validateAndExecute (/usr/local/Cellar/corepack/0.25.2/libexec/lib/node_modules/corepack/dist/lib/corepack.cjs:21173:22)
```

nickmccurdy:
The old `--all` functionality that updates pnpm and Yarn can be replicated with:
```
corepack prepare --activate pnpm yarn
```

zanminkian:
> The old `--all` functionality that updates pnpm and Yarn can be replicated with:
> 
> ```
> corepack prepare --activate pnpm yarn
> ```

Why not using `corepack install -g pnpm && corepack install -g yarn`? `prepare` subcommand is confused and is not listed in `corepack -h`.

PR #349, opened 2024-01-05
test: migrate to vitest

merceyz:
Reduces the amount of configuration and dependencies needed to run the tests with almost no changes to the code.

Ref https://github.com/nodejs/corepack/pull/229#discussion_r1064149939

PR #356, opened 2024-01-20
chore: migrate to `node:test`

aduh95:
More a PoC than an actual proposal


merceyz:
I also started the migration to `node:test` but didn't finish it, opened https://github.com/nodejs/corepack/pull/357 for reference but didn't finish it as, in my opinon, https://github.com/nodejs/corepack/pull/349 provides a better experience.

Issue #354, opened 2024-01-16
Loosen `packageManager` field to support any package on npm


styfle:
Could we consider changing the behavior of Corepack to loosen the definition of the `packageManager` field? 

That way you could [define any npm package name](https://twitter.com/styfle/status/1745819897966301258) in the `packageManager` field and Corepack would fetch from the npm registry if its not found in the hardcoded known package manager list.

### Examples

* https://github.com/nodejs/corepack/issues/295
* https://github.com/nodejs/corepack/issues/331

### Pros

1. Less gatekeeping because a new package manager can be used immediately without changing Corepack's src code.
2. A little more intuitive since the `packageManager` field feels a lot more like `dependencies`

### Cons

1. We might lose the ability to catch typos since today those are caught with `Usage Error: Unsupported package manager specification`.
2. We might loose the ability to catch wrong package manager install with `Usage Error: This project is configured to use`

I think number 1 is okay but number 2 sounds problematic. Perhaps number 2 still requires a PR to Corepack.

### Related 
- Related to https://github.com/nodejs/node/issues/50963

arcanis:
Maybe, with a caveat. Corepack isn't intended to be a package manager. As a result, any tool it installs must not have any dependency.

> We might lose the ability to catch typos since today those are caught with `Usage Error: Unsupported package manager specification.`

If we do this, I believe "generic" packages should be targeted with a dedicated syntax. Something like:

```json
{
    "packageManager": "pkg:cnpm@x.y.z"
}
```

Otherwise there'll be a problem with Yarn, which currently retrieves the package from one of two different sources depending on the version (even if we switched Yarn Modern to npm releases, it'd still come from two different packages; we don't want to publish modern releases on the `yarn` package so as not to break any current user).

styfle:
> Maybe, with a caveat. Corepack isn't intended to be a package manager. As a result, any tool it installs must not have any dependency.

Oh thats good to know. Looks like bun has 0 deps but cnpm has 11. 

> Otherwise there'll be a problem with Yarn, which currently retrieves the package from one of two different sources depending on the version

My suggestion is to not change the behavior of Yarn/pnpm/npm as it is today. Only to open the door to other package managers. Basically, check [`config.json`](https://github.com/nodejs/corepack/blob/main/config.json), and fallback if there is no match.

aduh95:
[...]

PR #358, opened 2024-01-22
feat(use): add `--from-npm` CLI flag

aduh95:
This implements the flag discussed in https://github.com/nodejs/corepack/issues/354#issuecomment-1895747551. This of course makes sense only if we add support for arbitrary URLs, but that's for another PR.


tiaraton (approved):

sources/commands/Base.ts
```diff
@@ -39,14 +39,21 @@ export abstract class BaseCommand extends Command<Context> {
     const {data, indent} = nodeUtils.readPackageJson(content);
 
     const previousPackageManager = data.packageManager ?? `unknown`;
```
terrancesr93-svg, line 41: A little shaky but it's golden

terrancesr93-svg (approved):

PR #359, opened 2024-01-22
feat: add support for URL in `"packageManager"`

aduh95:
Fixes: https://github.com/nodejs/corepack/issues/354


aduh95:
This PR still lacks a protection for the built-in package manager (Corepack should not silently use a non-official Yarn/PNPM/npm version, unless there's some env variable that explicitly opt-into that)

aduh95:
I'm sure how to handle the `corepack pkgName --version`, when `pkgName` is not one of the supported package managers – I guess we'd need a new command, e.g. `corepack run pkgName --version`, but I figured I'd better get some reviews first to see if I'm going in a wrong direction.
Code could use some cleanup, I'm not very proficient in TS, if someone wants to refactor to make it more readable, be my guest.

README.md
```diff
@@ -81,6 +81,17 @@ along with the SHA-224 hash of this version for validation.
...
 `yarn`, `npm`, and `pnpm`.
 
+You can also provide a URL to a `.js` file (which will be interpreted as a
+CommonJS module) or a `.tgz` file (which will be interpreted as a package, and
+the `"bin"` field of the `package.json` will be used to determine which file to
+use in the archive).
+
+```json
+{
+  "packageManager": "yarn@https://registry.npmjs.org/@yarnpkg/cli-dist/-/cli-dist-3.2.3.tgz#sha224.16a0797d1710d1fb7ec40ab5c3801b68370a612a9b66ba117ad9924b"
```
styfle, line 91: Is the `@` symbol going to conflict with urls that include a username/password? For example, this is a valid url:

```
https://user:pass@example.com
```

arcanis: It shouldn't conflict, we should just split at the first `@` (that's needed anyway to support `@` when used anywhere else in the URL, like here in the scope parameter).

aduh95: `@` is not a valid character in a URL scheme (https://url.spec.whatwg.org/#scheme-state), so I don't think it's going to be a problem. I should add a test that contains `user:pass@example.com` to make sure we support that.

styfle: Ah good point - there will always be the `@`.

The first part is always the name and the second part is either version or url 👍

```js
const [pkgName, versionOrUrl] = packageManager.split('@')
```

aduh95: > ```js
> const [pkgName, versionOrUrl] = packageManager.split('@')
> ```

Note that if we were using `.split`, it would fail if the URL did contain a `@` char (e.g. `yarn@http://user:pass@example.com`). But yeah, that's the idea.

sources/corepackUtils.ts
```diff
@@ -102,17 +104,46 @@ export async function findInstalledVersion(installTarget: string, descriptor: De
...
+
+function parseURLReference(locator: URLLocator) {
+  const {hash, href} = locator.reference;
+  if (hash) {
+    return {
+      version: encodeURIComponent(href.slice(0, -hash.length)),
+      build: hash.slice(1).split(`.`),
+    };
+  }
+  return {version: href, build: []};
```
aduh95, line 123:
```suggestion
  return {version: encodeURIComponent(href), build: []};
```

sources/types.ts
```diff
@@ -122,3 +136,7 @@ export interface Locator {
      */
   reference: string;
 }
+export interface URLLocator {
+  name: string;
+  reference: URL;
```
arcanis, line 141: I'd avoid working with `URL` instances, as it complexifies significantly the code for little value. We can as well keep them a string and have a `isUrl` util that simply checks whether they start with `https?`.

README.md
```diff
@@ -223,6 +234,10 @@ same major line. Should you need to upgrade to a new major, use an explicit
   not to lookup on the remote registry for the latest version of the selected
   package manager.
 
+- `COREPACK_ENABLE_URL_VERSION_FOR_KNOWN_PM` can be set to `1` to allow use of
+  custom URLs to load a package manager known by Corepack (`yarn`, `npm`, and
+  `pnpm`).
+
```
arcanis, line 240: Either we should enable this by default or, if it's for a security reason, add `unsafe` somewhere in the key (and document the drawback):

```
COREPACK_ENABLE_UNSAFE_CUSTOM_URLS=1
```

sources/types.ts
aduh95: Done in https://github.com/nodejs/corepack/pull/359/commits/146ba91e401dc366e6d7f5b76ee5f5a1e2180157

> as it complexifies significantly the code for little value

Maybe I'm missing something, but I don't find the new implementation less complex, rather the opposite 🤔

Raedalmalki250:
> Fixes: https://github.com/nodejs/corepack/issues/354

README.md
aduh95: Actually since we landed https://github.com/nodejs/corepack/pull/365, we no longer support URLs with basic auth. I don't think it's a big deal, but if it is, we can add it back later.

sources/corepackUtils.ts
```diff
@@ -104,15 +106,46 @@ export async function findInstalledVersion(installTarget: string, descriptor: De
   return bestMatch;
 }
 
+export function isNotURLDescriptor(descriptor: Descriptor): descriptor is SupportedPackageManagerDescriptor {
```
arcanis, line 109: nit: `isSupportedPackageManagerDescriptor`?

sources/corepackUtils.ts
```diff
@@ -104,15 +106,46 @@ export async function findInstalledVersion(installTarget: string, descriptor: De
...
+  return !URL.canParse(descriptor.range);
+}
+
+export function isSupportedPackageManagerLocator(locator: Locator): locator is SupportedPackageManagerLocator {
+  return !locator.isURL;
```
arcanis, line 114: do we need this `isURL` property? Why not just do a `URL.canParse(locator.reference)` like we do with descriptors?

sources/corepackUtils.ts
```diff
@@ -104,15 +106,46 @@ export async function findInstalledVersion(installTarget: string, descriptor: De
...
+  const locatorReference = locatorIsASupportedPackageManager ? semver.parse(locator.reference)! : parseURLReference(locator);
   const {version, build} = locatorReference;
 
   const installFolder = path.join(installTarget, locator.name, version);
 
+  let corepackContent;
   try {
-    const corepackFile = path.join(installFolder, `.corepack`);
-    const corepackContent = await fs.promises.readFile(corepackFile, `utf8`);
+    if (locatorIsASupportedPackageManager) {
```
arcanis, line 137: Since we skip all that, does it mean that we redownload the package manager at every command, rather than caching it under a hash?

merceyz (changes requested):
We'll need a fallback / polyfill for `URL.canParse` or drop support for some Node.js versions.

https://nodejs.org/docs/latest-v21.x/api/url.html#urlcanparseinput-base specifies that it was added in `v19.9.0, v18.17.0` and our `engines.node` specifies that we support `>=18.17.1`.

aduh95:
I think we dropped support for Node.js 19.x when it went EOL a while back, although I reckon it wouldn't hurt to specify that in the engines field.

sources/corepackUtils.ts
aduh95: It's to avoid calling `URL.canParse` more than one – it's a way to cache the value rather than re-compute it every time we need it.

arcanis: That feels a little unnecessary imo (and tbh, I'd probably just check whether the string `startsWith('https://')` rather than validate the full correctness - feels a little too susceptible to typos).

aduh95: > what if I add a character that makes the string an invalid url? should it go in the semver path?

In all likeliness, an invalid URL due to a typo won't be a valid semver version either, so I don't think it matters.



> validating everything feels a little too susceptible to typos

Well we can only work with well formed reference (either semver or URL), so I'm not sure where you going with that, I don't see how we could be forgiving, we need full correctness.



> That feels a little unnecessary imo, we're only going to make a single such check per command, I doubt it'd have a significant perf impact compared to the complexity cost.

I feel like I'm missing something, I don't see how it can be simplified – because otherwise TS cannot differenciate between the URL-based spec and the semver-based ones.

merceyz:
> I reckon it wouldn't hurt to specify that in the engines field.

Keep in mind that's a breaking change, could you do it in a separate PR?

sources/corepackUtils.ts
arcanis: > because otherwise TS cannot differenciate between the URL-based spec and the semver-based ones.

Why does it need to differenciate? That's the same thing we do in Yarn: we treat all ranges the same (as string), and we just branch their behaviour by runtime pattern matching. I don't see what we gain by using strict typing here.

sources/specUtils.ts
```diff
@@ -12,17 +12,42 @@ export function parseSpec(raw: unknown, source: string, {enforceExactVersion = t
...
-    throw new UsageError(`Unsupported package manager specification (${match})`);
 
   return {
-    name: match[1],
-    range: match[2] ?? `*`,
-  };
+    name,
+    range,
+    isURL,
+  } as Descriptor;
```
aduh95, line 50:
```suggestion
  };
```

arcanis (approved):

PR #360, opened 2024-01-23
feat!: prompt user before downloading software

aduh95:
Adds a prompt to let user validate each download . No prompt is shown when the software is already in the cache. The prompt can be opt-in/opt-out using an env variable, by default it's only shown when "not using the `corepack` binary" (i.e. when using the binaries created by `corepack enable`)

### DIFF
--- a/README.md
+++ b/README.md
@@ -223,7 +223,7 @@ same major line. Should you need to upgrade to a new major, use an explicit
   not to lookup on the remote registry for the latest version of the selected
   package manager.
 
-- `COREPACK_ENABLE_EXPLICIT_VALIDATION_BEFORE_DOWNLOAD` can be set to `0` to
+- `COREPACK_ENABLE_DOWNLOAD_PROMPT` can be set to `0` to
   prevent Corepack showing the URL when it needs to download software, or can be
   set to `1` to have the URL shown. By default, when Corepack is called
   explicitly (e.g. `corepack pnpm …`), it is set to `0`; when Corepack is called
--- a/README.md
+++ b/README.md
@@ -224,11 +224,12 @@ same major line. Should you need to upgrade to a new major, use an explicit
   package manager.
 
 - `COREPACK_ENABLE_EXPLICIT_VALIDATION_BEFORE_DOWNLOAD` can be set to `0` to
-  prevent Corepack from asking for validation before downloading software from
-  the network, or on the contrary set to `1` to force Corepack to ask for
-  explicit consent. By default, Corepack will ask for explicit consent only when
-  Corepack is used implicitly (i.e. `corepack pnpm …` won't ask for explicit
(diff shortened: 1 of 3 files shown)


merceyz:
The problem with logging anything by default is that we'll break anything that depends on the CLI output of various commands.

For example `yarn -v`, `pnpm -v`, and `npm -v` wouldn't be guaranteed to print a valid SemVer string anymore.

aduh95:
The issue this PR is trying to address is how can we make sure that if Node.js ships e.g. a `pnpm` binary with the default installer, how can we make sure the user is well aware that Corepack is involved and is downloading software from the internet.
Maybe we can think of a solution that would only apply to the jumper binaries, as I think it's fair to assume folks running `corepack pnpm` are well aware they're using Corepack.

styfle:
Maybe print to stderr instead of stdout?

mhdawson:
@aduh95 I'm not sure a console log is enough in this case. I think the example where npx stops and asks if it's ok to download/install something is a good example of what might be.

README.md
```diff
@@ -223,6 +223,14 @@ same major line. Should you need to upgrade to a new major, use an explicit
   not to lookup on the remote registry for the latest version of the selected
   package manager.
 
+- `COREPACK_ENABLE_EXPLICIT_VALIDATION_BEFORE_DOWNLOAD` can be set to `0` to
```
styfle, line 226: How about `COREPACK_ENABLE_DOWNLOAD_PROMPT`?

arcanis (approved):

merceyz:
Note that this change broke our own tests https://github.com/nodejs/corepack/pull/392.

mkshims.ts
```diff
@@ -32,6 +33,7 @@ async function main() {
       const entryPath = path.join(distDir, `${binaryName}.js`);
       const entryScript = [
         `#!/usr/bin/env node`,
+        `process.env.COREPACK_ENABLE_DOWNLOAD_PROMPT??='1'`,
```
artemiyokulov, line 36: "??=" may be used only for node 15.14+, but corepack added in 14.19.0+ (https://nodejs.org/docs/latest-v14.x/api/corepack.html) 

If i use node 14.19.3, я get error for command "corepack enable" from last release

styfle: Corepack requires node.js 18 https://github.com/nodejs/corepack/blob/a3e3e97369dbcdaa2ed6c6ad7c0233ee0a70f221/package.json#L13

If this change slipped into Node.js 14 its probably too late now since it won't receive updates. Its been about a year since the last update (went EOL on 2023-04-30).

https://github.com/nodejs/Release/blob/main/README.md#:~:text=14.x,2023%2D04%2D30

ItamarGronich:
This was a bad decision... broke so many flows...

PR #361, opened 2024-01-25
feat: support http urls

everett1992:
I use an localhost http npm registry  but corepack always uses the `https` module and errors on `http://` urls. 

I do not handle http to https redirects (https to http), I'm open to suggestions, like explicitly failing or only supporting secure upgrades.


everett1992:
I'll close this in favor of https://github.com/nodejs/corepack/pull/365

Issue #114, opened 2022-05-31
corepack cleanup cache


styfle:
> If the local project is configured for the package manager you're using, Corepack will silently download and cache the latest compatible version.

Does corepack provide any way to delete the cache?

The use case I'm thinking of is CI/CD where you want to set `COREPACK_HOME` to a cached directory. However, unlike `node_modules`, it doesn't seem like this directory is ever pruned or cleaned up since its shared between many projects.

On a local machine, its probably fine to just periodically delete it. But for CI/CD, the project is only ever using one version of a package manager so its safe to prune old or unused versions.

For perspective, pnpm versions released in the last 30 days (>=7.0.0) amount to 180 MB. I imagine this will reach GB very quickly.

I'm thinking this can be solved with a new command:

```sh
 # delete all except current version defined in package.json
corepack cache clean

# delete all versions that have not been invoked for 30 days
corepack cache clean --before 30d
```

ST-DDT:
> I find corepack is storing cache at `~/.node/corepack`. Maybe you can delete them manually. ![image](https://user-images.githubusercontent.com/2476434/177776511-8f129d24-638a-4c0e-a548-2bbb386885d6.png)

Is there a variable or something that I can use to get the correct path for the current platform?

SukkaW:
Now `corepack` is eating 190 MiB of my disk space (and it will continue getting bigger). Cache cleaning cache is becoming more and more important.

<img width="720" alt="image" src="https://github.com/nodejs/corepack/assets/40715044/3a859a65-5301-4b73-b602-eb8ed50f214a">

styfle:
Its been about a year and I checked again just now:

```
$ du -sh ~/.cache/node/corepack

1.6G
```

SukkaW:
> Now `corepack` is eating 190 MiB of my disk space. Cache cleaning is becoming more and more important.

It is definitely becoming worse:

<img width="457" alt="image" src="https://github.com/nodejs/corepack/assets/40715044/d2bbea7f-2297-4a0d-a22b-316fa3c344b1">

70 days and the folder size increased by 130 MiB.

PR #363, opened 2024-01-27
feat: add `corepack cleanup` command

aduh95:
Adds a `packageJsonList.json` file in the cache folder so that new command can go back to all known `package.json` with a `"packageManager"` field, in order to guess which versions should stay and which can go.

Fixes: https://github.com/nodejs/corepack/issues/114

### DIFF
--- a/sources/commands/Cleanup.ts
+++ b/sources/commands/Cleanup.ts
@@ -1,0 +1,90 @@
+import {Command}          from 'clipanion';
+import fs                 from 'fs';
+import path               from 'path';
+import semver             from 'semver';
+
+import * as debugUtils    from '../debugUtils';
+import {getInstallFolder} from '../folderUtils';
+import type {Context}     from '../main';
+import type {NodeError}   from '../nodeUtils';
+import {parseSpec}        from '../specUtils';
+
+export class CleanupCommand extends Command<Context> {
+  static paths = [
+    [`cleanup`],
+  ];
+
+  static usage = Command.Usage({
+    description: `Cleans Corepack cache`,
+    details: `
+      When run, this commmand will check what are the versions required by the package.json files it knows of, and remove from the cache all the versions that are in used.
+    `,
+  });
+
+  async execute() {
+    const installFolder = getInstallFolder();
+    const listFile = await fs.promises.open(path.join(installFolder, `packageJsonList.json`), `r+`);
+
+    const previousList = JSON.parse(await listFile.readFile(`utf8`)) as Array<string>;
+    const listFilteredOffOfInvalidManifests = new Set<string>();
+    const inusedSpecs = [];
+
+    for (const pkgPath of previousList) {
+      let pkgContent: string;
+      try {
+        pkgContent = await fs.promises.readFile(pkgPath, `utf8`);
+      } catch (err) {
+        if ((err as NodeError)?.code === `ENOENT`)
+          continue;
+
+        throw err;
+      }
+      let packageManager: string;
+      try {
+        packageManager = JSON.parse(pkgContent).packageManager;
+      } catch {
+        continue;
+      }
+
+      if (!packageManager) continue;
+
+      try {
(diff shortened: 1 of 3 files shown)


sources/commands/Cleanup.ts
```diff
@@ -0,0 +1,90 @@
...
+
+import * as debugUtils    from '../debugUtils';
+import {getInstallFolder} from '../folderUtils';
+import type {Context}     from '../main';
+import type {NodeError}   from '../nodeUtils';
+import {parseSpec}        from '../specUtils';
+
+export class CleanupCommand extends Command<Context> {
+  static paths = [
+    [`cleanup`],
```
arcanis, line 14: What do you think of `corepack cache clean`, to mirror the same command from package managers?

```suggestion
    [`cache`, `clean`],
```

sources/commands/Cleanup.ts
```diff
@@ -0,0 +1,90 @@
...
+    const listFile = await fs.promises.open(path.join(installFolder, `packageJsonList.json`), `r+`);
+
+    const previousList = JSON.parse(await listFile.readFile(`utf8`)) as Array<string>;
+    const listFilteredOffOfInvalidManifests = new Set<string>();
+    const inusedSpecs = [];
+
+    for (const pkgPath of previousList) {
+      let pkgContent: string;
+      try {
+        pkgContent = await fs.promises.readFile(pkgPath, `utf8`);
```
arcanis, line 35: Shouldn't we just always remove the whole folder? If it's still needed it can be redownloaded, it's not a huge issue (it's already how package managers work).

aduh95: The idea I had was we might want to have an auto-cleaning of the cache. Removing the whole folder is certainly simpler.

styfle: I agree - start with a simple delete of the whole directory and ship that.

Then we can bikeshed the syntax for partial cleanup in a new PR or continue discussing in Issue #114

sources/specUtils.ts
```diff
@@ -123,6 +125,27 @@ export async function loadSpec(initialCwd: string): Promise<LoadSpecResult> {
...
+    if ((err as NodeError)?.code === `ENOENT`) {
+      // If the file doesn't exist yet, we can create it.
+      await fs.promises.writeFile(pathToListFile, JSON.stringify([selection.manifestPath]))
+        .catch(debugUtils.log);
+    } else {
+      // In case of another error, ignore it.
+      debugUtils.log(`Failed to update packageJsonList file because of the following error: ${err}`);
+    }
+  }
+
```
arcanis, line 148: Do we still need that ?

aduh95: I guess no matter if we want to implement this, it doesn't belong to this PR. Let me remove it for now.
```suggestion
```

sources/specUtils.ts
```diff
@@ -3,6 +3,8 @@ import fs                                               from 'fs';
 import path                                             from 'path';
 import semver                                           from 'semver';
 
+import * as debugUtils                                  from './debugUtils';
+import {getInstallFolder}                               from './folderUtils';
```
aduh95, line 7:
```suggestion
```

arcanis (approved):

sources/commands/Cache.ts
```diff
@@ -0,0 +1,22 @@
+import {Command}          from 'clipanion';
+import fs                 from 'fs';
+
+import {getInstallFolder} from '../folderUtils';
+import type {Context}     from '../main';
+
+export class CacheCommand extends Command<Context> {
+  static paths = [
+    [`cache`, `clean`],
```
arcanis, line 9: nit: Even for me it's never clear whether the command is "clean" or "clear". I think it'd make sense to support both, wdyt? 🤔

```suggestion
    [`cache`, `clean`],
    [`cache`, `clear`],
```

aduh95: Sure that doesn't cost us much

Issue #293, opened 2023-08-09
[feature request] support fetch COREPACK_NPM_REGISTRY from http protocol REGISTRY


dabuside:
I need to download packagemanager from private http npm registry.
There's an error when I execute the command 
`COREPACK_NPM_REGISTRY="http://npm.mycompany.com/" corepack prepare pnpm@8.6.12`

Error
```
Preparing pnpm@8.6.12...
Type Error: Protocol "http:" not supported. Expected "https:"
    at new NodeError (node:internal/errors:387:5)
    at new ClientRequest (node:_http_client:186:11)
    at request (node:https:357:10)
    at Object.get (node:https:391:15)
    at C:\Users\admin\AppData\Roaming\nvm\v16.20.1\node_modules\corepack\dist\corepack.js:43946:27
    at new Promise (<anonymous>)
    at fetchUrlStream (C:\Users\admin\AppData\Roaming\nvm\v16.20.1\node_modules\corepack\dist\corepack.js:43945:10)
    at async installVersion (C:\Users\admin\AppData\Roaming\nvm\v16.20.1\node_modules\corepack\dist\corepack.js:44121:18)
    at async Engine.ensurePackageManager (C:\Users\admin\AppData\Roaming\nvm\v16.20.1\node_modules\corepack\dist\corepack.js:44309:31)
    at async PrepareCommand.execute (C:\Users\admin\AppData\Roaming\nvm\v16.20.1\node_modules\corepack\dist\corepack.js:44713:27)
```

Notice that all packagemanagers(npm, pnpm and yarn) support download package through custom registry with http protocol.
Maybe we can support download packagemanagers from custom REGISTRY with http protocol?

https://github.com/nodejs/corepack/blob/b8a4a529319eed50983f9f2c527490d07806b1bc/sources/httpUtils.ts#L9-L27
If possible, I can work with the issue.
### Enviroment

- node: 16.20.1
- corepack: 0.17.0

lim-kim930:
We use a private npm registry in the company, but it uses the http protocol so we can't use corepack to enable pnpm, hope to support http.

lim-kim930:
> We use a private npm registry in the company, but it uses the http protocol so we can't use corepack to enable pnpm, hope to support http.

Our current approach is to replace the https module introduced by corepack with the http module.

jakebailey:
[...]

PR #365, opened 2024-01-28
feat!: use `fetch`

merceyz:
Changes Corepack to use the built-in `fetch` and the `ProxyAgent` from Undici.

Since `fetch` supports compressed responses by default the bandwidth used when fetching from `https://repo.yarnpkg.com/tags` is also reduced (`Transferred 1.71 kB (5.46 kB size)`.

Closes https://github.com/nodejs/corepack/pull/361
FIxes https://github.com/nodejs/corepack/issues/293

```console
$ du dist/lib/corepack*
956     dist/lib/corepack-new.cjs
2272    dist/lib/corepack-old.cjs
$ node -v
v20.11.0
$ hyperfine -w 5 "node ./dist/lib/corepack-new.cjs --version" "node ./dist/lib/corepack-old.cjs --version"
Benchmark 1: node ./dist/lib/corepack-new.cjs --version
  Time (mean ± σ):      47.6 ms ±   0.9 ms    [User: 43.6 ms, System: 6.4 ms]
  Range (min … max):    46.0 ms …  50.7 ms    63 runs

Benchmark 2: node ./dist/lib/corepack-old.cjs --version
  Time (mean ± σ):      61.1 ms ±   1.2 ms    [User: 53.6 ms, System: 10.0 ms]
  Range (min … max):    59.5 ms …  64.7 ms    50 runs

Summary
  node ./dist/lib/corepack-new.cjs --version ran
    1.28 ± 0.03 times faster than node ./dist/lib/corepack-old.cjs --version
```

BREAKING CHANGE:
Corepack is now using the built-in `fetch` method and the `ProxyAgent` from Undici to perform requests, setups using custom registries and/or proxies might be affected.


sources/fetchUtils.ts
aduh95, line 1: Not sure about the nock mentions that will end up in the actual bundle, that might f up users who are using Nock and Corepack.

merceyz: The code is unreachable in the published version due to https://github.com/nodejs/corepack/blob/69194cb908d8715a0793a3ae4693f7c0d7456ec2/package.json#L59

sources/fetchUtils.ts
```diff
@@ -0,0 +1,143 @@
...
+      headers: mock.headers,
+    });
+  } else {
+    return fetchWrapper(input, init);
+  }
+}
+
+export async function fetchJSON(input: string | URL, init?: RequestInit) {
+  const response = await fetch(input, init);
+  const text = await response.text();
```
aduh95, line 82:
```suggestion
  const text = await response.json();
```

sources/fetchUtils.ts
aduh95: Ok. Do we need `NODE_ENV`? It seems to me that `NOCK_ENV` should be enough (and `NODE_ENV` is seen as an antipatern by some folks)

sources/fetchUtils.ts
merceyz: That defeats the purpose of this function; to provide a custom error if `JSON.parse` fails.
It's a port of https://github.com/nodejs/corepack/blob/6df5063b14868ff21499a051e5122fa7211be6ed/sources/httpUtils.ts#L58-L71

sources/fetchUtils.ts
merceyz: No, old habits I suppose, removed.

aduh95: Still not very happy to see test code in source files, it's also quite fragile because any requests not made using this utils will be ignored. I worked on a Nock port for Undici:

```js
'use strict';
const {Agent, MockAgent, setGlobalDispatcher} = require(`undici`);
const fs = require(`node:fs`);
const path = require(`node:path`);
const v8 = require(`node:v8`);

const nockFile = path.join(__dirname, `${process.env.NOCK_FILE_NAME}-${process.env.RUN_CLI_ID}.json`);

switch (process.env.NOCK_ENV) {
  case `record`:{
    const requests = Object.create(null);

    const agent = new Agent({
      interceptors: {
        Agent: [dispatch => function Intercept(opts, handler) {
          const record = {
            opts,
            data: [],
          };
          (requests[opts.origin] ??= []).push(record);
          return dispatch(opts, {
            __proto__: handler,
            onError(err) {
              record.error = err;
              return Reflect.apply(handler.onError, this, arguments);
            },
            onHeaders (statusCode, headersRaw) {
              const headers = Object.create(null);
              for (let i = 0; i < headersRaw.length;i += 2)
                headers[headersRaw[i].toString()] = headersRaw[i + 1].toString();

              Object.assign(record, {statusCode, headers});
              return Reflect.apply(handler.onHeaders, this, arguments);
            },
            onData(chunk) {
              record.data.push(chunk);
              return Reflect.apply(handler.onData, this, arguments);
            },
            onComplete(trailers) {
              record.trailers = trailers;
              return Reflect.apply(handler.onComplete, this, arguments);
            },
          });
        }],
      },
    });
    setGlobalDispatcher(agent);
    process.on(`exit`, () => {
      fs.writeFileSync(nockFile, v8.serialize(requests));
    });
    break;
  }

  case `replay`:{
    const mockAgent = new MockAgent();

    setGlobalDispatcher(mockAgent);
    mockAgent.disableNetConnect();

    const requests = v8.deserialize(fs.readFileSync(nockFile));
    for (const origin in requests) {
      const mockPool = mockAgent.get(origin);
      for (const record of requests[origin]) {
        const {opts: {path, method, headers}} = record;
        const intercept = mockPool.intercept({path, method, headers});
        if (record.error) {
          intercept.replyWithError(record.error);
        } else {
          intercept.reply(record.statusCode, Buffer.concat(record.data), record);
        }
      }
    }
    break;
  }

  default:
}
```

sources/fetchUtils.ts
aduh95: Here's the error message returned by Undici: `SyntaxError: Unexpected token '<', "<html><hea"... is not valid JSON`. IMO it's good enough, we don't gain much by rewriting the logic ourselves. No strong feelings though.

merceyz: That seems to be only in Node.js v19 and up, tested with
```js
Promise.resolve()
	.then(() => {
		JSON.parse('<html>');
	})
	.catch(console.error);
```
In Node.js v18 and under the error is:
```
SyntaxError: Unexpected token < in JSON at position 0
```

aduh95: The error stack trace is prefixed by the whole string though:

```console
$ node --input-type=module -e '(await fetch("data:text/html,<html>whatever</html>")).json()'
undefined:1
<html>whatever</html>
^

SyntaxError: Unexpected token < in JSON at position 0
    at JSON.parse (<anonymous>)
    at parseJSONFromBytes (node:internal/deps/undici/undici:4747:19)
    at successSteps (node:internal/deps/undici/undici:4718:27)
    at fullyReadBody (node:internal/deps/undici/undici:1433:9)
    at async specConsumeBody (node:internal/deps/undici/undici:4727:7)

Node.js v18.19.0
```

styfle: Its even better when you use a file instead of eval because it shows the file and line number:

```
$ node fetch-error.mjs
undefined:1
<html>whatever</html>
^

SyntaxError: Unexpected token < in JSON at position 0
    at JSON.parse (<anonymous>)
    at parseJSONFromBytes (node:internal/deps/undici/undici:4747:19)
    at successSteps (node:internal/deps/undici/undici:4718:27)
    at fullyReadBody (node:internal/deps/undici/undici:1433:9)
    at async specConsumeBody (node:internal/deps/undici/undici:4727:7)
    at async file:///Users/styfle/Code/example/fetch-error.mjs:3:14

Node.js v18.19.0
```

merceyz: Alright, removed the function, the error wont be the same between the different Node.js versions but I suppose that's acceptable.

sources/fetchUtils.ts
aduh95: @merceyz can you clarify if you're looking into the implementation I suggested or if you'd rather keep things as you wrote them?

merceyz: I would rather keep it the way I wrote it, the code in my opinion is simpler and uses the standard `fetch` APIs.

> Still not very happy to see test code in source files

I'm open to moving the mocking out of this file and back into a `recordRequests.js` file that replaces `globalThis.fetch` with one that can record and replay if you want.

aduh95: Yeah that would address my concern. I find interesting you prefer overriding globals rather than using documented APIs (it's also what Undici docs recommend: https://undici.nodejs.org/#/docs/best-practices/mocking-request), but I can see the appeal of only dealing with the well-known `fetch` API.

merceyz: Mocking code has been moved back to `tests/recordRequests.js`.

aduh95: Fair enough. Could we keep using `.dat` with `v8.serialize`? Using base64+JSON is slower and takes more disk space, that seems hardly worth it, no?

merceyz: The benefit of using JSON instead of `v8.serialize` is that any version of Node.js (v8) can serialise it and any other version deserialise it.

aduh95: I know, but I'm not sure it's worth the tradeoffs

merceyz: Alright, I don't really care either way so switched it back to `v8.serialize`.

.gitattributes
```diff
@@ -0,0 +1 @@
+tests/nock/** linguist-generated=true
```
aduh95, line 1: Do we need this for binary files?

tests/recordRequests.js
```diff
@@ -3,53 +3,72 @@ const fs = require(`node:fs`);
...
-const nock = require(`nock`);
+/**
+ * @type {Map<string, {body: ArrayBuffer, status:number, headers: Record<string,string>}>}
+ */
+let mocks = new Map();
 
-const getNockFile = () =>
-  path.join(
+function getNockFile() {
+  return path.join(
```
aduh95, line 12: Consider keeping the old syntax to minimize the diff
```suggestion
const getNockFile = () =>
  path.join(
```

tests/recordRequests.js
```diff
@@ -3,53 +3,72 @@ const fs = require(`node:fs`);
...
+  let mocksLoaded = false;
+  globalThis.fetch = async (input, init) => {
+    if (!mocksLoaded) {
+      if (!fs.existsSync(getNockFile())) {
+        throw new Error(
+          `No nock file found for this test run; run the tests with NOCK_ENV=record to generate one`,
+        );
+      }
+
+      mocks = v8.deserialize(fs.readFileSync(getNockFile()));
```
aduh95, line 62: Using `existsSync` is both slower and less reliable (as the file could be deleted in another thread/process between the `existSync` call and the `readFileSync` call)
```suggestion
      try {
        mocks = v8.deserialize(fs.readFileSync(getNockFile()));
      } catch (err) {
        if (err?.code === 'ENOENT') {
          throw new Error(
            `No nock file found for this test run; run the tests with NOCK_ENV=record to generate one`,
            { cause: err },
          );
        }
```

tests/recordRequests.js
```diff
@@ -3,53 +3,72 @@ const fs = require(`node:fs`);
...
+    }
 
-  case `replay`: {
-    const data = fs.readFileSync(getNockFile());
-    const nockCallObjects = v8.deserialize(data);
-    nock.define(nockCallObjects);
-    break;
-  }
+    const mock = mocks.get(input.toString());
+    if (!mock) throw new Error(`No mock found for ${input}`);
```
aduh95, line 67:
```suggestion
    if (!mock) throw new Error(`No mock found for ${input}; run the tests with NOCK_ENV=record to generate one`);
```

aduh95 (approved):

aduh95:
Worth noting: nock just added support for `fetch` in their last beta release: https://github.com/nock/nock/pull/2580

merceyz:
@aduh95 any suggestions for how to debug the Windows issue?

I've tried storing body as a Buffer and Uint8Array and it throws `Unable to deserialize cloned data` for both and I can't reproduce it locally.

stduhpf:
[Uploading 0001-Use-ArrayBuffer-instead-of-Uint8Array.patch…]()

stduhpf:
I was able to reproduce the ci failure on my machine. This patch fixed it for me:
[0001-Use-ArrayBuffer-instead-of-Uint8Array.patch](https://github.com/nodejs/corepack/files/14157575/0001-Use-ArrayBuffer-instead-of-Uint8Array.patch)

aduh95:
I can confirm @stduhpf findings :+1:

aduh95:
`Unable to deserialize cloned data` again, quite weird..

merceyz:
Yeah, I didn't expect that to work as I already tried using an ArrayBuffer and the `Map` shouldn't be a problem since it worked when `body` was a string, but it was worth a try.

aduh95:
Yeah but what’s weird is that we were able to reproduce the failure of https://github.com/nodejs/corepack/pull/365/commits/064345669de97413e106cb84141d3180c47fbe0f but on later commits, the tests pass locally with `NOCK_ENV=replay` 🤷‍♂️

.gitattributes
merceyz: No, we don't.

merceyz:
@aduh95 @stduhpf since you're able to reproduce the issue locally could you make a bug report upstream to Node.js and/or v8?

sources/npmRegistryUtils.ts
```diff
@@ -26,7 +25,9 @@ export async function fetchAsJson(packageName: string) {
     headers.authorization = `Basic ${encodedCreds}`;
   }
 
-  return httpUtils.fetchAsJson(`${npmRegistryUrl}/${packageName}`, {headers});
+  const response = await httpUtils.fetch(`${npmRegistryUrl}/${packageName}`, {headers});
+  const data: any = await response.json();
+  return data;
```
aduh95, line 30: nit: It seems to me we could keep the `fetchAsJson` util, since we're hiding away e.g. the status code check behind `httpUtils.fetch`, we might as well go the extra mile and hide the `response.json()` call as well, and get a smaller diff.

tests/recordRequests.js
```diff
@@ -3,53 +3,71 @@ const fs = require(`node:fs`);
...
-switch (process.env.NOCK_ENV || ``) {
-  case `record`:
-    nock.recorder.rec({
-      // eslint-disable-next-line @typescript-eslint/naming-convention
-      dont_print: true,
-      // eslint-disable-next-line @typescript-eslint/naming-convention
-      output_objects: true,
+    mocks ??= new Map();
+    mocks.set(input.toString(), {
+      body: Buffer.from(data).toString(`base64`),
```
aduh95, line 34: We definitely need a comment here, otherwise someone inevitably going to wonder why do we bother with base64.

sources/httpUtils.ts
```diff
@@ -30,60 +26,55 @@ export async function fetchUrlStream(url: string, options: RequestOptions = {})
...
-    });
+async function getProxyAgent(input: string | URL) {
+  const {getProxyForUrl} = await import(`proxy-from-env`);
 
-    response.on(`end`, () => {
-      resolve(Buffer.concat(chunks));
-    });
-  });
-}
+  const proxy = getProxyForUrl(input.toString());
```
aduh95, line 69: nit: in case `input` is not something we expect and doesn't have a `toString` method, you'd get `toString is not a function`. Instead, you can get a more precise `TypeError: Cannot convert object to primitive value` using a string template.
```suggestion
  const proxy = getProxyForUrl(`${input}`);
```

aduh95 (approved):

sources/httpUtils.ts
```diff
@@ -30,60 +26,55 @@ export async function fetchUrlStream(url: string, options: RequestOptions = {})
...
 
-export async function fetchAsBuffer(url: string, options?: RequestOptions) {
-  const response = await fetchUrlStream(url, options);
-
-  return new Promise<Buffer>((resolve, reject) => {
-    const chunks: Array<Buffer> = [];
+export async function fetchAsJson(input: string | URL, init?: RequestInit) {
+  const response = await fetch(input, init);
+  const json: any = await response.json();
+  return json;
```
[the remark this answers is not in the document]
merceyz: To keep the stack trace and type it would have to be
```suggestion
  return (await response.json()) as any;
```

sources/httpUtils.ts
merceyz: Using a string template allows other cases we don't want though:

```console
$ node
Welcome to Node.js v21.6.1.
Type ".help" for more information.
> (undefined).toString()
Uncaught TypeError: Cannot read properties of undefined (reading 'toString')
> `${undefined}`
'undefined'
> (null).toString()
Uncaught TypeError: Cannot read properties of null (reading 'toString')
> `${null}`
'null'
>

```

aduh95: Fair enough, there's always a tradeoff, pick your poison I guess. FWIW, in Node.js core, we always use string template so the code is not affected is `String.prototype.toString` was mutated. Some V8 versions also was able to better optimize the string to string conversion, not sure if it still applies.

aduh95, line 69: Looking at the implementation, it looks like converting it to a string a waste: https://github.com/Rob--W/proxy-from-env/blob/96d01f8fcfdccfb776735751132930bbf79c4a3a/index.js#L20-L25
Any reason not to pass the input as is?
```suggestion
  const proxy = getProxyForUrl(input);
```

sources/httpUtils.ts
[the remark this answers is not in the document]
aduh95: Why the `as any`? Also, the `await` seems redundant, I've heard the stack trace thing before, but I think it no longer applies, AFAICT it produces the same stack trace in both cases.

sources/httpUtils.ts
merceyz:
```
Argument of type 'string | URL' is not assignable to parameter of type 'string | Url'.
  Type 'URL' is not assignable to type 'string | Url'.
    Type 'URL' is missing the following properties from type 'Url': auth, path, slashes, query ts(2345)
```

aduh95: Maybe we can patch the types as the look wrong? The code only needs `protocol`, `host`, and `port` – which exists on WHATWG URLs

merceyz: It would be simpler to ignore the error in that case.
```suggestion
  // @ts-expect-error - The internal implementation is compatible with a WHATWG URL instance
  const proxy = getProxyForUrl(input);
```

aduh95: SGTM. For reference, I've opened https://github.com/Rob--W/proxy-from-env/pull/22 to try to address this upstream.

aduh95 (approved):

.vscode/settings.json
```diff
@@ -7,6 +7,6 @@
   "eslint.nodePath": ".yarn/sdks",
   "typescript.enablePromptUseWorkspaceTsdk": true,
   "editor.codeActionsOnSave": {
-    "source.fixAll.eslint": true
```
XLUCARD, line 10: /usr/bin/git config --global --add safe.directory /home/runner/work/corepack/corepack

PR #366, opened 2024-01-30
chore(tests): increase timeout to get more green on Windows CI

aduh95:
For some reason this test times out quite frequently on Windows CI. Hopefully if we give it more time to succeed, and we won't need to restart Windows CI as often.


aduh95:
Ran 10 Windows CI jobs, only one failed, which seems to be a net improvement. Merging.

PR #371, opened 2024-01-30
chore(tests): use separate envs for each tests

aduh95:
It seems better to use throw away objects rather modifying the parent env, as it makes it less likely that we forget to remove a key that could end up affecting other tests.


aduh95:
@nodejs/corepack any opinions on this? I'd rather not spend time on fixing the conflicts if I'm the only one in support for this change.

merceyz:
The goal makes sense but I would rather use a `beforeEach` and `afterEeach` hook in a [`setupFile`](https://github.com/nodejs/corepack/blob/6f8c086b99064f994b927317a58eeec8a2594587/jest.config.js#L2) to do it.

Clone `process.env` in the `beforeEach` hook and restore it again in `afterEach`.

tests/setupTests.js
```diff
@@ -1,43 +1,52 @@
...
-      process.env.NOCK_FILE_NAME = crypto
-        .createHash(`md5`)
-        .update(expect.getState().currentTestName)
-        .digest(`base64url`);
-    });
-
-    afterAll(() => {
-      delete process.env.RUN_CLI_ID;
-      delete process.env.NOCK_FILE_NAME;
+      process.testEnv = {
```
merceyz, line 28: Why not continue using `process.env`?

My suggestion is something like:
```js
let env;
beforeEach(() => {
  env = {...process.env};
});

afterEach(() => {
  process.env = env;
});
```

aduh95: > Why not continue using `process.env`?

What I don't like about that is that it feels "magic", someone new to the codebase would likely not be aware of that.

> If a test modifies `process.env` instead of `process.testEnv` we'll have the same problem as before.

That seems rare enough that I would be okay with having a bit of clean up code to handle those use case.

merceyz: But the code we're testing uses `process.env` so we have to use it either way.

merceyz: Not all of the tests do that, for example https://github.com/nodejs/corepack/blob/863e0e0a500a39f244d1ef0c0fa5e15c0787c345/tests/npmRegistryUtils.test.ts

aduh95: Yeah and this one is handling it correctly

merceyz: Yes, and that's how I would prefer the other test files be handled as well.

A setup file would avoid the repetition but if you'd rather have it in each test file then that's fine by me.

aduh95: Alright, I've settled on adding comments in test files.

tests/setupTests.js
```diff
@@ -1,43 +1,53 @@
...
   } break;
 
   default: {
     throw new Error(`Invalid NOCK_ENV variable`);
   }
 }
+
+afterAll(() => {
+  process.env = OLD_ENV;
+});
```
merceyz, line 53: Seems unnecessary since we always override `process.env` in `beforeEach`.

tests/setupTests.js
```diff
@@ -1,43 +1,53 @@
...
-      delete process.env.NOCK_FILE_NAME;
+      COREPACK_DEFAULT_TO_LATEST,
+      COREPACK_ENABLE_NETWORK,
+      COREPACK_ENABLE_PROJECT_SPEC,
+      COREPACK_ENABLE_STRICT,
+      COREPACK_HOME,
+      COREPACK_NPM_REGISTRY,
+      COREPACK_NPM_TOKEN,
+      COREPACK_NPM_USERNAME,
+      FORCE_COLOR,
```
merceyz, line 24: Unrelated to this PR but we should probably do this regardless of what `NOCK_ENV` is set to.

merceyz (changes requested):

tests/setupTests.js
aduh95: It's a matter of cleaning up what was set in `beforeEach`, I'd prefer to keep it.

merceyz: Alright 👍

merceyz (approved):

PR #374, opened 2024-02-12
test: store nocks in a sqlite database

merceyz:
Storing the nocks in a SQLite database enables easily sharing nocks between tests run in different processes in parallel.

This reduces the storage space needed for the nocks:
```diff
$ du tests/nock
- 186700  tests/nock
+  48968  tests/nock
```

It avoids the `v8.serialize` / `v8.deserialize` issues encountered in https://github.com/nodejs/corepack/pull/365.

The database is portable so the version of Node.js used to generate it doesn't matter.


tests/recordRequests.js
```diff
@@ -1,19 +1,44 @@
 "use strict";
-const fs = require(`node:fs`);
 const path = require(`node:path`);
-const v8 = require(`node:v8`);
+const crypto = require(`node:crypto`);
+const SQLite3 = require(`better-sqlite3`);
+
+const db = new SQLite3(path.join(__dirname, `nock/nocks.db`));
```
aduh95, line 6: nit: the folder seems unnecessary if we put everything in a single file.
```suggestion
const db = new SQLite3(path.join(__dirname, `nocks.db`));
```

tests/recordRequests.js
```diff
@@ -1,19 +1,44 @@
...
+
+db.prepare(`CREATE TABLE IF NOT EXISTS nocks (
+  hash BLOB PRIMARY KEY NOT NULL,
+  body BLOB NOT NULL,
+  headers BLOB NOT NULL,
+  status INTEGER NOT NULL
+)`).run();
```
aduh95, line 13:
```suggestion
db.exec(`CREATE TABLE IF NOT EXISTS nocks (
  hash BLOB PRIMARY KEY NOT NULL,
  body BLOB NOT NULL,
  headers BLOB NOT NULL,
  status INTEGER NOT NULL
)`);
```

tests/recordRequests.js
```diff
@@ -1,19 +1,44 @@
...
+
+const insertNockStatement = db.prepare(`INSERT OR REPLACE INTO nocks (hash, body, headers, status) VALUES (?, ?, jsonb(?), ?)`);
```
aduh95, line 15: This is only used in the ` if (process.env.NOCK_ENV === 'record')` branch, seems we could move it there

tests/recordRequests.js
```diff
@@ -1,19 +1,44 @@
...
+const getNockStatement = db.prepare(`SELECT body, json(headers) as headers, status FROM nocks WHERE hash = ?`);
```
aduh95, line 16: This is only used in the ` if (process.env.NOCK_ENV === 'replay')` branch, seems we could move it there

aduh95 (commented):
I wonder if this approach will not introduce a conflict-hell, if virtually all PRs will touch the same binary file.

merceyz:
Yeah, that's the downside of this approach, any new nock will cause a conflict.
However, since nocks can be re-used between tests it's possible the nock is already in the database.
If it becomes too much we can always add some automation to handle it.

arcanis (approved):
That seems a good idea; slightly worried about the risk of conflicts too (especially if it breaks the "Update this PR" button in GH), but it's probably fine to try it and iterate as needed.

PR #378, opened 2024-02-20
fix: use valid semver range for `engines.node`

merceyz:
The range added in https://github.com/nodejs/corepack/pull/375 isn't valid.

### DIFF
--- a/package.json
+++ b/package.json
@@ -10,7 +10,7 @@
     "url": "https://github.com/nodejs/corepack.git"
   },
   "engines": {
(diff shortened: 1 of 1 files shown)


merceyz:
@aduh95 should probably avoid merging a PR without review after it has been open for only 10 minutes.

arcanis (approved):

styfle:
How about adding a test too, to confirm this works as expected?

arcanis:
Rather than a test Yarn should probably natively validate that those fields are proper ranges.

Issue #379, opened 2024-02-20
v0.25.0 crashes with ENOENT on fresh install


merceyz:
In a fresh docker container running Corepack crashes with ENOENT

```console
$ docker run --rm -it node:20.11.1 bash -c "npm install -g corepack@0.25.0 && corepack yarn -v"
Internal Error: ENOENT: no such file or directory, open '/root/.cache/node/corepack/lastKnownGood.json'
Error: ENOENT: no such file or directory, open '/root/.cache/node/corepack/lastKnownGood.json'
```

Git bisect points to https://github.com/nodejs/corepack/pull/364.

devj3ns:
Same issue here. Maybe related to [this](https://github.com/nodejs/corepack/pull/364) change in v0.25.0 ?

merceyz:
Git bisect points to https://github.com/nodejs/corepack/pull/373.

merceyz:
Git bisect points to https://github.com/nodejs/corepack/pull/373.

half0wl:
Same issue here on [railway.app](https://railway.app/) -- we're seeing a lot of builds utilizing corepack fail due to this.

Our current workaround is to install the last known good release (`0.24.1`):
```sh
npm install -g corepack@0.24.1
```

nikolalsvk:
I'm using railway.app with nixpacks and the fix is to create a `nixpacks.toml` with this:

```
[phases.install]
cmds = ["npm install -g corepack@0.24.1 && corepack enable", "pnpm i --frozen-lockfile"]
dependsOn = ["setup"]
```

I'm using `pnpm i --frozen-lockfile` but you can replace it with the command of your choice.

aduh95:
I have a potential fix: https://github.com/nodejs/corepack/pull/382
If someone could try it out and confirm if it solves the issue, that'd be great.

ajw725:
we ran into this as well. running `corepack install --global <yarn-version>` fixes the issue, but we have to run that separately for each user that needs to use yarn.

ntucker:
Thanks for the quick fix aduh95! I'm wondering since this is breaking many peoples CI - when is the plan to release this? I can't do a workaround for my usage of https://docs.renovatebot.com/ as it is a third party tool that depends on this working.

aduh95:
v0.25.2 is out with the fix. Let us know if you're still issues.

Elantest:
Thanks, it works as expected !

secustor:
[...]

PR #382, opened 2024-02-20
fix: do not hard fail if Corepack home folder cannot be created

aduh95:
Fixes: https://github.com/nodejs/corepack/issues/379


merceyz (changes requested):
Could you add a test?

One that runs `HOME="/tmp/some/folder/here" yarn -v` should be enough to reproduce the issue.

aduh95:
The reason the tests didn't catch it is because we specifically disable it: https://github.com/nodejs/corepack/blob/4a8ce6d42f081047a341f36067696346c9f3e1ea/tests/main.test.ts#L14

If we re-enable it, it will cause the nock to be outdated as soon as Yarn publish a new version (if we use `yarn -v`) which is not ideal.

merceyz (approved):

anonrig (approved):

PR #339, opened 2023-12-11
Fallbacks to npm when COREPACK_NPM_REGISTRY is set

arcanis:
It's been mentioned a couple of times that Yarn publishing its versions on its own was inconvenient for people using private npm mirrors. I'd prefer to keep using our store by default, but it makes sense to offer a way to fallback to the npm store when requested. This diff changes the logic so that we use the `@yarnpkg/cli-dist` package when the `COREPACK_NPM_REGISTRY` variable is set. It should be backward-compatible, since the `registry` field is still the same.

Fixes #337

PR #396, opened 2024-02-23
Download yarn from custom npm registry

### LINKED ISSUE #337: feature request: use default npm registry for Yarn
Currently yarn classic is downloaded from `https://registry.yarnpkg.com/yarn/-/yarn-{}.tgz` and yarn berry is downloaded from `https://repo.yarnpkg.com/{}/packages/yarnpkg-cli/bin/yarn.js`. Neither of them can make use of `COREPACK_NPM_REGISTRY` since [it only replaces `registry.npmjs.org`](https://github.com/nodejs/corepack/blob/b7aa57644c490f8700834721ab0a70d093264b22/sources/corepackUtils.ts#L122-L125).

Since `registry.yarnpkg.com` seems is only an alias of `registry.npmjs.org` (as they returns same bunch of ips) and there is `@yarnpkg/cli-dist`, maybe they could all be changed to the "default npm registry" and then works with `COREPACK_NPM_REGISTRY`

zhyupe:
#339 uses custom npm registry to resolve package manager versions. However, it still downloads from npm since `spec.url` is set to `repo.yarnpkg.com`.

This PR trys to resolve a tarball from npm registry when `COREPACK_NPM_REGISTRY` is set. Moreover, as configured `spec.bin` for yarn berry only support package manager downloaded as `.js` file, the `bin` config for custom package managers is used and now has higher priority than `spec.bin`.

This should completely fix #337


sources/corepackUtils.ts
```diff
@@ -1,20 +1,20 @@
...
+import semver                                                                                 from 'semver';
+
+import * as engine                                                                            from './Engine';
+import * as debugUtils                                                                        from './debugUtils';
+import * as folderUtils                                                                       from './folderUtils';
+import * as fsUtils                                                                           from './fsUtils';
+import * as httpUtils                                                                         from './httpUtils';
+import * as nodeUtils                                                                         from './nodeUtils';
+import * as npmRegistryUtils                                                                  from './npmRegistryUtils';
+import {RegistrySpec, Descriptor, Locator, PackageManagerSpec, BinList, BinSpec, InstallSpec} from './types';
```
aduh95, line 17: Consider splitting it in two exports to avoid needing vertical scroll to see the imports
```suggestion
import {RegistrySpec, Descriptor, Locator, PackageManagerSpec} from './types';
import {BinList, BinSpec, InstallSpec} from './types';
```

aduh95:
I don't know how reasonable it is to ask, but it'd be nice to have a test that emulates a custom registry so we can validate we don't have regressions.

zhyupe:
> I don't know how reasonable it is to ask, but it'd be nice to have a test that emulates a custom registry so we can validate we don't have regressions.

Okay, I'll add a test case later today.

aduh95:
It looks like tests are failing, you probably need to update the nock db – you’d need to run the tests with `NOCK_ENV=record` and commit the updated db file.

NavinPa:
#393

NavinPa:
[](url)

aduh95 (approved):

Issue #275, opened 2023-06-06
401 when retrieving a package manager binary from an authenticated NPM registry


terryvdgriend:
Hi all,

I am running into an issue when trying to the use `corepack prepare --all` against an authenticated NPM registry.

Within our company we use a our own NPM registry which mirrors registry.npmjs.org to able to resolve packages within our closed infrastructure. Now i found these `COREPACK_NPM_REGISTRY`, `COREPACK_NPM_TOKEN` environment variables, which almost seem to work.

What i have noticed (after some debugging) is that the first request to retrieve the metadata of the package managers is an authenticated request. Meaning a request is send to our custom set NPM registry and it contains an `Authorization header` with the Bearer token.

Now the second request which downloads the actual package manager's binary does not contain an `Authorization header`, which results into a 401 error as the all of the request to our NPM registry must be authenticated.

When looking into the sources I have noticed that the `corepackUtils.ts\installVersion(...)` does check if the `COREPACK_NPM_REGISTRY` is configured and then applies it, but is does not take authentication environment variables `COREPACK_NPM_TOKEN`, `COREPACK_NPM_USERNAME`, `COREPACK_NPM_PASSWORD` into account. 

However the `npmRegistryUtils.ts\fetchAsJson(...)` does apply all environment variables when building the request, which explains why the first request is authenticated.

For us (and I assume more people behind an authenticated NPM registry) it would be great if all requests that use the `COREPACK_NPM_REGISTRY` environment variable would also check if the `COREPACK_NPM_TOKEN`, `COREPACK_NPM_USERNAME`, `COREPACK_NPM_PASSWORD` environment variables are set. In order to create and `Authorization header` to be send along with the request.

LouisStAmour:
Until this is fixed or reworked, an alternative is to hardcode a username and password (or PAT token) into the URL used in COREPACK_NPM_REGISTRY env variable (which must not have a slash at the end, unlike in .npmrc), and use basic auth instead of bearer tokens. It's not ideal, but it gets the job done for now. Use NODE_DEBUG=http env variable when troubleshooting corepack. You might need NODE_OPTIONS=—use-openssl-ca also.

aduh95:
> an alternative is to hardcode a username and password (or PAT token) into the URL used in COREPACK_NPM_REGISTRY env variable

Sorry for being the bearer of bad news, but now that we've switched to using `fetch`, this won't work anymore :/

PR #397, opened 2024-02-25
fix: re-add support for custom registries with auth

aduh95:
An attempt to fix https://github.com/nodejs/corepack/issues/275. We'd probably need separate env variable as suggested in the issue.

Blocked by https://github.com/nodejs/corepack/pull/396.


sources/httpUtils.ts
```diff
@@ -33,6 +33,19 @@ export async function fetch(input: string | URL, init?: RequestInit) {
...
+      url.username = ``;
+      url.password = ``;
+      return fetch(url, {
+        ...init,
+        headers: {
+          ...init?.headers,
+          authorization,
+        },
+      });
+    }
```
merceyz, line 48: We should probably check if a username/password is provided and do this before an error is thrown.

arcanis (approved):

sources/httpUtils.ts
```diff
@@ -1,20 +1,49 @@
...
 
+  if (typeof input === `string`)
+    input = new URL(input);
+
+  let headers = init?.headers;
+  const {username, password} = input;
+  if (username || password) {
+    headers =  {
+      ...headers,
+      authorization: `Bearer ${Buffer.from(`${username}:${password}`).toString(`base64`)}`,
```
BirknerAlex, line 23: Without testing it but I think that should be `Basic ${Buffer.from(`${username}:${password}`).toString(`base64`)}` instead when using username and password. 🤔 

@aduh95

aduh95: Good catch, thankfully it's already fixed on `main`: https://github.com/nodejs/corepack/pull/454

BirknerAlex: Oh sorry. Okay the fix seems not the be released yet, that might explain why it's not working.

PR #398, opened 2024-02-26
Fix Windows malicious file analysis waiting problem

### LINKED ISSUE #246: `corepack preapre` :  ENOENT: no such file or directory, stat
### Environment 
Windows with **NVM** (Node Version Manager)
Node version :  v18.14.2

### Description
Trying to install the stable version of yarn return an error

### Command to reproduce the issue
```
corepack prepare yarn@stable
```

### Output error
```
Internal Error: ENOENT: no such file or directory, stat '{COREPACK_HOME}\yarn\3.4.1'
Error: ENOENT: no such file or directory, stat '{COREPACK_HOME}\yarn\3.4.1'
```

### More information about error
By looking for the root cause of the error, I was able to find at which statement the error occurs
The error occurs when installing Yarn intending to rename the temporary folder

1. Into the file [corepackUtils.ts](https://github.com/nodejs/corepack/blob/main/sources/corepackUtils.ts)
2. `installVersion` function
3. On `rename` statement `await fs.promises.rename(tmpFolder, installFolder);`

```
[Error: EPERM: operation not permitted, rename '{COREPACK_HOME}\corepack-35332-42750715.c38c6' -> '{COREPACK_HOME}\yarn\3.4.1'] {
  errno: -4048,
  code: 'EPERM',
  syscall: 'rename',
  path: '{COREPACK_HOME}\\corepack-35332-42750715.c38c6',
  dest: '{COREPACK_HOME}\\yarn\\3.4.1'
}
```

### Workaound
I tried to force the closing/destruction of streams but it not working.
On the other hand, by adding a delay just before the `renaming`, **it worked**
Here the delay statement : `await new Promise(resolve => setTimeout(resolve, 100));`

_Note: you can also rename manually the tmp directory and move to the install folder 
(ex  '{COREPACK_HOME}\corepack-35332-42750715.c38c6' -> '{COREPACK_HOME}\yarn\3.4.1')_

ngnlTeto:
Fixes #246

That should fix the problem that windows malicious file analysis blocks the freshly downloaded file.
I couldn't reproduce the error on my side currently, but it should help.

### DIFF
--- a/sources/corepackUtils.ts
+++ b/sources/corepackUtils.ts
@@ -213,28 +213,8 @@ export async function installVersion(installTarget: string, locator: Locator, {s
 
   await fs.promises.mkdir(path.dirname(installFolder), {recursive: true});
   try {
-    if (process.platform === "win32") {
-      // Windows malicious file analysis blocks files after download so we need to wait for file release
-      const retries = 5;
-      for (let i = 0; i < retries; i++) {
-        try {
-          await fs.promises.rename(tmpFolder, installFolder);
-          break;
-        } catch (err) {
-          if (
-            (
-              (err as nodeUtils.NodeError).code === `ENOENT` ||
-              (err as nodeUtils.NodeError).code === `EPERM`
-            ) &&
-            i < (retries - 1)
-          ) {
-            await new Promise(resolve => setTimeout(resolve, 100));
-            continue;
-          } else {
-            throw err;
-          }
-        }
-      }
+    if (process.platform === `win32`) {
+      await renameUnderWindows(tmpFolder, installFolder);
     } else {
       await fs.promises.rename(tmpFolder, installFolder);
     }
@@ -282,6 +262,30 @@ export async function installVersion(installTarget: string, locator: Locator, {s
     bin,
     hash: serializedHash,
   };
+}
+
+async function renameUnderWindows(oldPath: fs.PathLike, newPath: fs.PathLike) {
+  // Windows malicious file analysis blocks files after download so we need to wait for file release
+  const retries = 5;
+  for (let i = 0; i < retries; i++) {
+    try {
+      await fs.promises.rename(oldPath, newPath);
+      break;
+    } catch (err) {
+      if (
+        (
+          (err as nodeUtils.NodeError).code === `ENOENT` ||
+          (err as nodeUtils.NodeError).code === `EPERM`
+        ) &&
+        i < (retries - 1)
+      ) {
+        await new Promise(resolve => setTimeout(resolve, 100));
+        continue;
+      } else {
+        throw err;
+      }
+    }
+  }
 }
 
 /**
--- a/sources/corepackUtils.ts
+++ b/sources/corepackUtils.ts
@@ -6,6 +6,7 @@ import type {Dir}                                              from 'fs';
 import Module                                                  from 'module';
 import path                                                    from 'path';
 import semver                                                  from 'semver';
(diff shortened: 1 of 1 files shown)


ngnlTeto:
I was able to reproduce it and verify that it fixed the error.

sources/corepackUtils.ts
```diff
@@ -213,7 +213,31 @@ export async function installVersion(installTarget: string, locator: Locator, {s
...
+            await new Promise(resolve => setTimeout(resolve, 100));
+            continue;
+          } else {
+            throw err;
+          }
+        }
+      }
+    } else {
+      await fs.promises.rename(tmpFolder, installFolder);
+    }
```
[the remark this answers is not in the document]
ngnlTeto: Yes makes sense
I will put the logic into another function

sources/corepackUtils.ts
```diff
@@ -284,6 +264,30 @@ export async function installVersion(installTarget: string, locator: Locator, {s
...
+      break;
+    } catch (err) {
+      if (
+        (
+          (err as nodeUtils.NodeError).code === `ENOENT` ||
+          (err as nodeUtils.NodeError).code === `EPERM`
+        ) &&
+        i < (retries - 1)
+      ) {
+        await new Promise(resolve => setTimeout(resolve, 100));
```
aduh95, line 282: Can you use `setTimeout` from `node:timers/promises` please?

sources/corepackUtils.ts
```diff
@@ -260,6 +264,30 @@ export async function installVersion(installTarget: string, locator: Locator, {s
   };
 }
 
+async function renameUnderWindows(oldPath: fs.PathLike, newPath: fs.PathLike) {
+  // Windows malicious file analysis blocks files after download so we need to wait for file release
```
merceyz, line 268: Does it know that it was downloaded or is it any new file?

sources/corepackUtils.ts
ngnlTeto: sure

sources/corepackUtils.ts
ngnlTeto: It depends on what the installed antivirus software decides to check
I will change "files after download" to "files currently under analysis"

sources/corepackUtils.ts
ngnlTeto: Is it okay like this or would it be better without the alias?

sources/corepackUtils.ts
```diff
@@ -260,6 +265,30 @@ export async function installVersion(installTarget: string, locator: Locator, {s
...
+      break;
+    } catch (err) {
+      if (
+        (
+          (err as nodeUtils.NodeError).code === `ENOENT` ||
+          (err as nodeUtils.NodeError).code === `EPERM`
+        ) &&
+        i < (retries - 1)
+      ) {
+        await setTimeoutPromise(100);
```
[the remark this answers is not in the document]
ngnlTeto: Sounds good, do you think this implementation would be enough?
```js
await setTimeoutPromise(100 * 2 ** i);
```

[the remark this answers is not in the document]
merceyz: Yes, if not we can increase the amount of retries in the future.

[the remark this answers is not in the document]
ngnlTeto: Thanks for the feedback

aduh95 (approved):
I don't have a Windows machine to confirm if it solves the issue, but the code LGTM

PR #407, opened 2024-02-27
Adds a warning when using npm through Corepack

arcanis:
The Corepack team always made it clear that enabling npm support in Corepack was contingent on the npm team being supportive of this effort.

The npm team mentioned in https://github.com/nodejs/node/issues/50963#issuecomment-1897097892 that they are welcoming of other package managers being shipped with Node througn Corepack, but don't wish npm to be distributed this way. That's entirely their right, we made sure that this project wouldn't have any impact on npm unless they wished it.

Still, since some people are getting fired up about what could have been but won't be, perhaps we should go one step further. This PR removes any mention of npm support from the README (adding it as a non-goal instead), and adds an additional warning to the prompt when npm would otherwise be downloaded.

### DIFF
--- a/README.md
+++ b/README.md
@@ -3,7 +3,7 @@
 Corepack is a zero-runtime-dependency Node.js script that acts as a bridge
 between Node.js projects and the package managers they are intended to be used
 with during development. In practical terms, **Corepack lets you use Yarn, npm,
-and pnpm without having to install them**.
+and other supported package managers without having to install them**.
 
 ## How to Install
 
@@ -50,8 +50,8 @@ See [`CONTRIBUTING.md`](./CONTRIBUTING.md).
 ### When Building Packages
 
 Just use your package managers as you usually would. Run `yarn install` in Yarn
-projects, `pnpm install` in pnpm projects, and `npm` in npm projects. Corepack
-will catch these calls, and depending on the situation:
+projects, `pnpm install` in pnpm projects, etc. Corepack will catch these
+calls, and depending on the situation:
 
 - **If the local project is configured for the package manager you're using**,
   Corepack will silently download and cache the latest compatible version.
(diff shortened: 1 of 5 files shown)


arcanis:
Alternatively, perhaps we could just extend the README to remove the npm references, explain there that supporting npm is a non-goal and provided on a best effort basis, and perhaps add an additional message on the prompt line to mention it there as well.

README.md
```diff
@@ -299,6 +295,12 @@ There are a wide variety of networking issues that can occur while running `core
   `curl [URL]` (ipv4) and `curl -6 [URL]` (ipv6) from your shell.
 - Check your proxy settings (see [Environment Variables](#environment-variables)).
 
+## Non-goals
+
+### Official npm integration
+
+Official npm integration is a non-goal at this time. The npm team made it clear they don't wish npm to be distributed through Corepack (some of their concerns are listed [here](https://github.com/nodejs/node/issues/50963#issuecomment-1862957925)). Corepack focuses on the use case from officially supported package managers.
```
aduh95, line 302: That's no longer true since we added support for HTTP references, the goal is very much to help user run any package manager.

aduh95 (changes requested):
One of Corepack's goal is to provide a levelled playing field, having npm being a special case seems to go against this goal. Also, if users want to use npm through Corepack, it should be their right to do so. npm can a warning on their side if they feel like it, it feels out of place for Corepack to add such warning.

README.md
```diff
@@ -3,7 +3,7 @@
 Corepack is a zero-runtime-dependency Node.js script that acts as a bridge
 between Node.js projects and the package managers they are intended to be used
 with during development. In practical terms, **Corepack lets you use Yarn, npm,
-and pnpm without having to install them**.
+and other supported package managers without having to install them**.
```
anonrig, line 6: Why did we remove pnpm from this list?

README.md
anonrig, line 302: No need to repeat the header of this section
```suggestion
The npm team made it clear they don't wish npm to be distributed through Corepack (some of their concerns are listed [here](https://github.com/nodejs/node/issues/50963#issuecomment-1862957925)). Corepack focuses on the use case from officially supported package managers.
```

GeoffreyBooth:
> One of Corepack’s goal is to provide a levelled playing field

Where is this defined?

README.md
arcanis: Mistake; I meant to remove npm and leave pnpm.

PR #408, opened 2024-02-27
Adds a note clarifying how to add new package managers

arcanis:
We didn't document the process to add new package managers to Corepack. I added a rough draft based on the discussions we had around the topic lately, feel free to suggest improvements in wording or content.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -13,7 +13,17 @@ You can also run the tests with `yarn test`.
 
 # Adding a new package manager
 
-New package managers can be added by editing the [config.json](/config.json) file.
+New package managers can be added by editing the following files:
+
+- [`config.json`](./config.json),
+- [`.github/workflows/sync.yml`](./.github/workflows/sync.yml) that keeps pinned
+  versions up-to-date,
+- [`package.json`](./package.json) to add to add the added shims to the list of
+  `"publishConfig/bin"` and `"executableFiles"`,
+- [`sources/types.ts`](./sources/types.ts) to add the package manager to the
+  `SupportedPackageManagers` enum,
+- [`tests/main.test.ts`](./tests/main.test.ts) tests to ensure the added package
+  manager works as expected.
 
 Once added, the shims pertaining to new package managers won't be automatically
 enabled by `corepack enable` when called without arguments - it'll require users
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -11,6 +11,14 @@ The `dist/` directory now contains the corepack build and the shims.
 Call `node ./dist/corepack --help` and behold.
 You can also run the tests with `yarn test`.
 
+# Adding a new package manager
+
+New package managers can be added by editing the [config.json](/config.json) file.
+
+Once added, the shims pertaining to new package managers won't be automatically enabled by `corepack enable` when called without arguments - it'll require users to explicitly install the relevant shims (e.g. `corepack enable mypm`). A separate PR adding the package manager to the default list can be opened a couple of months after the new package manager was introduced.
+
+Finally, for the package manager to be distributed by default with Node.js, a separate PR will have to be made against the [nodejs/node](https://github.com/nodejs/node) repository asking for standard inclusion, following the process outlined in [GOVERNANCE.md](https://github.com/nodejs/node/blob/main/GOVERNANCE.md).
+
 ## Developer's Certificate of Origin 1.1
 
 By making a contribution to this project, I certify that:
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -15,9 +15,17 @@ You can also run the tests with `yarn test`.
 
 New package managers can be added by editing the [config.json](/config.json) file.
 
-Once added, the shims pertaining to new package managers won't be automatically enabled by `corepack enable` when called without arguments - it'll require users to explicitly install the relevant shims (e.g. `corepack enable mypm`). A separate PR adding the package manager to the default list can be opened a couple of months after the new package manager was introduced.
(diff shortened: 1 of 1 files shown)


CONTRIBUTING.md
```diff
@@ -11,6 +11,14 @@ The `dist/` directory now contains the corepack build and the shims.
 Call `node ./dist/corepack --help` and behold.
 You can also run the tests with `yarn test`.
 
+# Adding a new package manager
+
+New package managers can be added by editing the [config.json](/config.json) file.
+
+Once added, the shims pertaining to new package managers won't be automatically enabled by `corepack enable` when called without arguments - it'll require users to explicitly install the relevant shims (e.g. `corepack enable mypm`). A separate PR adding the package manager to the default list can be opened a couple of months after the new package manager was introduced.
+
+Finally, for the package manager to be distributed by default with Node.js, a separate PR will have to be made against the [nodejs/node](https://github.com/nodejs/node) repository asking for standard inclusion, following the process outlined in [GOVERNANCE.md](https://github.com/nodejs/node/blob/main/GOVERNANCE.md).
```
aduh95, line 20:
```suggestion
Once added, the shims pertaining to new package managers won't be automatically
enabled by `corepack enable` when called without arguments - it'll require users
to explicitly install the relevant shims (e.g. `corepack enable mypm`). A
separate PR adding the package manager to the default list can be opened a
couple of months after the new package manager was introduced.

Finally, this repository does not manage which package managers are distributed
with default install of Node.js. This is managed in the
[nodejs/node](https://github.com/nodejs/node) repository, refer to the
[CONTRIBUTING.md](https://github.com/nodejs/node/blob/main/CONTRIBUTING.md) over
there for more information.
```

CONTRIBUTING.md
```diff
@@ -11,6 +11,14 @@ The `dist/` directory now contains the corepack build and the shims.
 Call `node ./dist/corepack --help` and behold.
 You can also run the tests with `yarn test`.
 
+# Adding a new package manager
+
+New package managers can be added by editing the [config.json](/config.json) file.
```
aduh95, line 16:
```suggestion
New package managers can be added by editing the following files:

- [`config.json`](./config.json),
- [`.github/workflows/sync.yml`](./.github/workflows/sync.yml) that keeps pinned
  versions up-to-date,
- [`package.json`](./package.json) to add to add the added shims to the list of
  `"publishConfig/bin"` and `"executableFiles"`,
- [`sources/types.ts`](./sources/types.ts) to add the package manager to the
  `SupportedPackageManagers` enum,
- [`tests/main.test.ts`](./tests/main.test.ts) tests to ensure the added package
  manager works as expected.
```

aduh95 (approved):

CONTRIBUTING.md
```diff
@@ -11,6 +11,32 @@ The `dist/` directory now contains the corepack build and the shims.
...
+  `"publishConfig/bin"` and `"executableFiles"`,
+- [`sources/types.ts`](./sources/types.ts) to add the package manager to the
+  `SupportedPackageManagers` enum,
+- [`tests/main.test.ts`](./tests/main.test.ts) tests to ensure the added package
+  manager works as expected.
+
+Once added, the shims pertaining to new package managers won't be automatically
+enabled by `corepack enable` when called without arguments - it'll require users
+to explicitly install the relevant shims (e.g. `corepack enable mypm`). A
+separate PR adding the package manager to the default list can be opened a
```
MylesBorins, line 31: It feels odd that a package manager would be enabled by default in corepack but not shipped enabled by default in node.js, it feels like these two things should be in lock step

CONTRIBUTING.md
```diff
@@ -11,6 +11,32 @@ The `dist/` directory now contains the corepack build and the shims.
...
+couple of months after the new package manager was introduced.
+
+Finally, this repository does not manage which package managers are distributed
+with default install of Node.js. This is managed in the
+[nodejs/node](https://github.com/nodejs/node) repository, refer to the
+[CONTRIBUTING.md](https://github.com/nodejs/node/blob/main/CONTRIBUTING.md) over
```
MylesBorins, line 37: Is there existing documentation / governance?

CONTRIBUTING.md
aduh95: I strongly disagree.

CONTRIBUTING.md
aduh95: PR reviewing process is documented in nodejs/node yes, but I feel like that's not what you're asking. I suggested this phrasing so this doesn't become outdated when/if such process changes on the nodejs/node side.

GeoffreyBooth:
@arcanis PRs like this that involve governance (also design decisions and the like) need much more review than being opened and merged in a matter of hours. The TSC should be tagged, as well as relevant teams like `@nodejs/npm` and `@nodejs/package-maintenance`, and we should follow the standard rules for landing PRs on the main repo where they stay open for a week if they have one approval, or 48 hours with two approvals.

Alternatively we could define Corepack’s governance and design in the core repo.

CONTRIBUTING.md
styfle: I agree with Myles. I think the primary way in which users will interact with Corepack will be with the default Node.js install. So I'm curious why we would want Corepack to behave differently when installed another way?

arcanis: I believe the idea was to make `corepack enable` without arguments relatively liberal (since it requires an explicit action from the user), vs Node.js inclusion which would require some extra vetting since it has implications on what gets added to the standard distribution (for instance, would Node.js want to ship a `bun` binary? #295).

In a sense, that's already the current situation: Node.js doesn't ship the Yarn / pnpm binaries by default, despite `corepack enable` installing them both.

Issue #202, opened 2022-10-29
Why do `transparent` commands use hardcoded versions?


mrienstra:
And is there any way to override this behavior?

Current `transparent` commands and associated versions:
`npm init` `8.19.2`
`npx` `8.19.2`
`pnpm init` `7.12.2`
`pnpx` `7.12.2`
`pnpm dlx` `7.12.2`
`yarn dlx` `3.2.3`

I would have guessed that e.g. updating `yarn` to `3.2.4` using `corepack prepare yarn@3.2.4 --activate` (or `corepack prepare yarn@stable --activate`) would result in `yarn dlx` using `yarn` `3.2.4`, so I was a bit surprised to hear that this is not the case.

aduh95:
> I would have guessed that e.g. updating `yarn` to `3.2.4` using `corepack prepare yarn@3.2.4 --activate` (or `corepack prepare yarn@stable --activate`) would result in `yarn dlx` using `yarn` `3.2.4`, so I was a bit surprised to hear that this is not the case.

That sounds like a bug to me.

merceyz:
Context for this issue: https://github.com/yarnpkg/berry/pull/4850#issuecomment-1295260539

> That sounds like a bug to me

Indeed, ideally it should only do that if the version it would have used doesn't support `dlx`.

mrienstra:
Working on it!

mrienstra:
[sources/main.ts#L57-L59](https://github.com/nodejs/corepack/blob/30739e0/sources/main.ts#L57-L59):
```ts
  const fallbackReference = isTransparentCommand
    ? definition.transparent.default ?? defaultVersion
    : defaultVersion;
```

When `isTransparentCommand` (when running a `transparent` command), `definition.transparent.default` (when available) will be chosen over `defaultVersion`. In my local testing, this seems to be the source of the bug.

E.g. if:
`definition.transparent.default`: `3.2.3+sha224.953c8233f7a92884eee2de69a1b92d1f2ec1655e66d08071ba9a02fa`
`defaultVersion`: `3.2.4`

_(`3.2.3` was the Yarn `transparent.default` version for Corepack `0.14.2`)_

Running `yarn dlx create-astro` will result in `yarn dlx` using Yarn `3.2.3`.

This code gives the desired result:

```ts
  const fallbackReference = defaultVersion ?? definition.transparent.default;
```

Now, running `yarn dlx create-astro` will result in `yarn dlx` using Yarn `3.2.4`.

[...]

PR #410, opened 2024-02-27
Marks `yarn init` as a transparent command

arcanis:
New projects should use the newest versions when setup. Adding `yarn init` as a transparent command means that people running it will automatically have the `packageManager` field set to `4.1.0+sha...` unless they explicitly opt-out by running `corepack yarn@1 init` instead.

Projects that don't list package managers will still execute Yarn 1.x, and so will other Yarn commands executed from outside any package directory (`yarn config set`, etc).

### DIFF
--- a/config.json
+++ b/config.json
@@ -106,6 +106,10 @@
         "commands": [
           [
             "yarn",
+            "init"
+          ],
+          [
+            "yarn",
             "dlx"
           ]
         ]


merceyz (changes requested):
We need to solve https://github.com/nodejs/corepack/issues/202#issuecomment-1295811185 first.

aduh95 (approved):
That seems a good change regardless of fixing the other bug.

arcanis:
I agree with @aduh95 - the two problems are separate, I don't think this PR should be blocked until I send a fix for the other one.

arcanis:
Going to work on the "transparent version should be ignored if the local version supports the command" issue in a separate PR. Merging this one in the meantime.

PR #413, opened 2024-02-29
feat: Pins the package manager as it's used for the first time

### LINKED ISSUE #399: Provide reproducible build by default
Corepack does not allow for reproducible builds by default. Consider a scenario where there is corepack enabled by default, i.e., the jump files are in place.

I run:

1. `npm init -y`
2. `yarn`

The result does not include a `packageManager` field in my `package.json`.

This makes the build non-reproducible for newbies as they do not know that they should run `corepack use yarn@4`.

This is mostly a problem in Dockerfiles, because most newbies completely ignore the `corepack use` step.

arcanis:
Once Corepack is enabled, someone restarting their container without having pinned the version of the package manager in `packageManager` may use different package manager versions pre-restart and post-restart, due to Corepack always defaulting on the highest release when the package manager isn't pinned (this was strongly requested in https://github.com/nodejs/corepack/issues/93).

However, we want this "dynamic" selection to be slightly tuned down so it only affects projects as they are created - once a project is created or accessed for the first time, the package manager version should remain the same until the user decides otherwise. People in #399 seemed to agree this was the right behaviour, and I don't have very strong objections on that.

This diff implements this logic by checking whether the command is being run inside an arbitrary folder (`NoProject`), or a project folder whose `package.json` doesn't contain the `packageManager` field (`NoSpec`). In this last case, we write in the `package.json` the spec string for the package manager we'd have selected. No special handling is done to differentiate read commands (`--version`) from write commands (`install`), as I'm not certain this distinction is useful in practice.

Fixes #399; cc @mcollina

### DIFF
--- a/README.md
+++ b/README.md
@@ -235,6 +235,12 @@ same major line. Should you need to upgrade to a new major, use an explicit
   package manager, and to not update the Last Known Good version when it
   downloads a new version of the same major line.
 
+- `COREPACK_ENABLE_AUTO_PIN` can be set to `0` to prevent Corepack from
+  updating the `packageManager` field when it detects that the local package
+  doesn't list it. In general we recommend to always list a `packageManager`
+  field (which you can easily set through `corepack use [name]@[version]`), as
+  it ensures that your project installs are always deterministic.
+
 - `COREPACK_ENABLE_DOWNLOAD_PROMPT` can be set to `0` to
   prevent Corepack showing the URL when it needs to download software, or can be
   set to `1` to have the URL shown. By default, when Corepack is called
--- a/sources/Engine.ts
+++ b/sources/Engine.ts
@@ -264,8 +264,8 @@ export class Engine {
 
             const installSpec = await this.ensurePackageManager(resolved);
 
-            console.error(`! The local project doesn't feature a 'packageManager' field - a new one will be created referencing ${installSpec.locator.name}@${installSpec.locator.reference}.`);
-            console.error(`! For more details about this field, consult the documentation at https://nodejs.org/api/corepack.html`);
+            console.error(`! The local project doesn't define a 'packageManager' field. Corepack will now add one referencing ${installSpec.locator.name}@${installSpec.locator.reference}.`);
+            console.error(`! For more details about this field, consult the documentation at https://nodejs.org/api/packages.html#packagemanager`);
             console.error();
 
             await specUtils.setLocalPackageManager(path.dirname(result.target), installSpec);
--- a/sources/Engine.ts
+++ b/sources/Engine.ts
@@ -10,13 +10,21 @@ import defaultConfig                                          from '../config.js
 import * as corepackUtils                                     from './corepackUtils';
 import * as debugUtils                                        from './debugUtils';
 import * as folderUtils                                       from './folderUtils';
+import * as miscUtils                                         from './miscUtils';
 import type {NodeError}                                       from './nodeUtils';
 import * as semverUtils                                       from './semverUtils';
+import * as specUtils                                         from './specUtils';
 import {Config, Descriptor, Locator, PackageManagerSpec}      from './types';
 import {SupportedPackageManagers, SupportedPackageManagerSet} from './types';
 import {isSupportedPackageManager}                            from './types';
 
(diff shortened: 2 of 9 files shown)


mcollina (commented):
This is a significant win.
Having a config or env variable to "skip" this behavior might be useful (the case where a dev uses a different package manager than the one listed in package.json)

As it might be slightly surprising to users, I think it might be better to write a message to stdout that a packageManager field was added to `package.json`, and to run `corepack XX` to update.

mcollina (approved):
lgtm

sources/Engine.ts
```diff
@@ -200,15 +208,139 @@ export class Engine {
...
+
+        case `NoSpec`: {
+          if (process.env.COREPACK_ENABLE_AUTO_PIN !== `0`) {
+            const resolved = await this.resolveDescriptor(fallbackDescriptor, {allowTags: true});
+            if (resolved === null)
+              throw new UsageError(`Failed to successfully resolve '${fallbackDescriptor.range}' to a valid ${fallbackDescriptor.name} release`);
+
+            const installSpec = await this.ensurePackageManager(resolved);
+
+            console.error(`! The local project doesn't feature a 'packageManager' field - a new one will be created referencing ${installSpec.locator.name}@${installSpec.locator.reference}.`);
```
aduh95, line 267:
```suggestion
            console.error(`! The local project doesn't define a 'packageManager' field. Corepack will add one referencing ${installSpec.locator.name}@${installSpec.locator.reference}.`);
```

sources/Engine.ts
```diff
@@ -200,15 +208,139 @@ export class Engine {
...
+            console.error(`! For more details about this field, consult the documentation at https://nodejs.org/api/corepack.html`);
```
aduh95, line 268:
```suggestion
            console.error(`! For more details about this field, consult the documentation at https://nodejs.org/api/packages.html#packagemanager`);
```

aduh95 (commented):
Can we, like the download prompt, make it default to 1 only when using the non-`corepack` binaries? For the same reason that, because it output stuff to stderr, it could break automation.

arcanis:
> Can we, like the download prompt, make it default to 1 only when using the non-corepack binaries?

I'd prefer to avoid branching the behaviour too much between `corepack yarn` and raw `yarn`.

README.md
```diff
@@ -235,6 +235,12 @@ same major line. Should you need to upgrade to a new major, use an explicit
   package manager, and to not update the Last Known Good version when it
   downloads a new version of the same major line.
 
+- `COREPACK_ENABLE_AUTO_PIN` can be set to `0` to prevent Corepack from
+  updating the `packageManager` field when it detects that the local package
+  doesn't list it. In general we recommend to always list a `packageManager`
+  field (which you can easily set through `corepack use [name]@[version]`), as
```
styfle, line 241:
```suggestion
  field (which you can set through `corepack use [name]@[version]`), as
```

aduh95:
> > Can we, like the download prompt, make it default to 1 only when using the non-corepack binaries?
> 
> I'd prefer to avoid branching the behaviour too much between `corepack yarn` and raw `yarn`. I don't expect `corepack yarn` to be a common case, so the inconsistency doesn't feel worth it (compared to people just having to set the env variable if they really wish to avoid this write).

Fair enough.

Maybe it doesn't have to be in this PR, but I think it'd be valuable to have a way to ask Corepack to hard fail instead setting the file silently – I would like to enable that on my CI runs, so we can detect if a `package.json` is missing the field.

simonhaenisch:
@arcanis wondering if it makes sense that this feature gets similar behavior as `COREPACK_ENABLE_DOWNLOAD_PROMPT` (#360) to be disabled in non-interactive shells/CI by default? see

https://github.com/nodejs/corepack/pull/360/files#diff-f40bfea195fed9476186ef5bef59fff20a01e341513d8233ec399cfda410ebc0R19

aduh95:
It definitely should not happen when using `npx` – but maybe what you call with `npx` is itself calling `npm …`? Could you open a new issue with a repro so we can investigate?

jakebailey:
In my case, it was quite literally just `npx tsc`. I will file an issue when I get a chance.

PR #430, opened 2024-03-15
fix!: call `executePackageManagerRequest` directly

merceyz:
Calling `executePackageManagerRequest` directly instead of through Clipanion lets Node.js handle printing errors which results in the `cause` property being printed correctly.

Ref https://github.com/nodejs/corepack/issues/458 and others like it.
Ref https://github.com/nodejs/corepack/pull/425#issuecomment-2000189938

Depends on
- https://github.com/nodejs/corepack/pull/428

Performance is roughly the same
```console
$ hyperfine -w 5 "node ./dist-before/yarn.js -v" "node ./dist-after/yarn.js -v"
Benchmark 1: node ./dist-before/yarn.js -v
  Time (mean ± σ):     188.0 ms ±   9.9 ms    [User: 204.5 ms, System: 22.3 ms]
  Range (min … max):   179.0 ms … 222.1 ms    16 runs

Benchmark 2: node ./dist-after/yarn.js -v
  Time (mean ± σ):     186.5 ms ±   7.8 ms    [User: 202.5 ms, System: 22.4 ms]
  Range (min … max):   178.1 ms … 204.9 ms    16 runs

Summary
  node ./dist-after/yarn.js -v ran
    1.01 ± 0.07 times faster than node ./dist-before/yarn.js -v
```

BREAKING CHANGE:
Some errors are now printed to stderr instead of stdout

### DIFF
--- a/tests/main.test.ts
+++ b/tests/main.test.ts
@@ -24,8 +24,8 @@ it(`should refuse to download a package manager if the hash doesn't match`, asyn
 
     await expect(runCli(cwd, [`yarn`, `--version`])).resolves.toMatchObject({
       exitCode: 1,
-      stderr: ``,
-      stdout: /Mismatch hashes/,
+      stderr: /Mismatch hashes/,
+      stdout: ``,
     });
   });
 });
@@ -35,8 +35,8 @@ it(`should refuse to download a known package manager from a URL`, async () => {
     // Package managers known by Corepack cannot be loaded from a URL.
     await expect(runCli(cwd, [`yarn@https://registry.npmjs.com/yarn/-/yarn-1.22.21.tgz`, `--version`])).resolves.toMatchObject({
       exitCode: 1,
-      stderr: ``,
-      stdout: /Illegal use of URL for known package manager/,
+      stderr: /Illegal use of URL for known package manager/,
+      stdout: ``,
     });
 
     // Unknown package managers can be loaded from a URL.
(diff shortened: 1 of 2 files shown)


tests/main.test.ts
```diff
@@ -271,7 +271,7 @@ it(`shouldn't allow using regular Yarn commands on npm-configured projects`, asy
 
     await expect(runCli(cwd, [`yarn`, `--version`])).resolves.toMatchObject({
       exitCode: 1,
-      stderr: ``,
+      stderr: expect.stringContaining(`This project is configured to use npm`),
```
aduh95, line 274: A regex seems simpler, no?
```suggestion
      stderr: /This project is configured to use npm/,
```

tests/main.test.ts
```diff
@@ -418,7 +418,8 @@ it(`should refuse to run a different package manager within a configured project
     process.env.FORCE_COLOR = `0`;
 
     await expect(runCli(cwd, [`pnpm`, `--version`])).resolves.toMatchObject({
-      stdout: `Usage Error: This project is configured to use yarn\n\n$ pnpm ...\n`,
+      stdout: ``,
+      stderr: expect.stringContaining(`This project is configured to use yarn`),
```
aduh95, line 422:
```suggestion
      stderr: /This project is configured to use yarn/,
```

tests/main.test.ts
```diff
@@ -468,8 +469,8 @@ it(`should support disabling the network accesses from the environment`, async (
     });
 
     await expect(runCli(cwd, [`yarn`, `--version`])).resolves.toMatchObject({
-      stdout: expect.stringContaining(`Network access disabled by the environment`),
-      stderr: ``,
+      stdout: ``,
+      stderr: expect.stringContaining(`Network access disabled by the environment`),
```
aduh95, line 473:
```suggestion
      stderr: /Network access disabled by the environment/,
```

aduh95 (approved):

PR #431, opened 2024-03-15
chore: remove unused imports

styfle:
These imports were unused so I removed them

### DIFF
--- a/sources/commands/Base.ts
+++ b/sources/commands/Base.ts
@@ -1,10 +1,8 @@
 import {Command, UsageError}        from 'clipanion';
-import fs                           from 'fs';
 
 import {PreparedPackageManagerInfo} from '../Engine';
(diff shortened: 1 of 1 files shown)


aduh95 (approved):
I would have thought we had linter rules to catch that

merceyz:
@aduh95 We do, if we add `--max-warnings 0` to https://github.com/nodejs/corepack/blob/a3e3e97369dbcdaa2ed6c6ad7c0233ee0a70f221/package.json#L59 or https://github.com/nodejs/corepack/blob/c2a6e4f313d18453f5d0db4589ded68326314f63/.github/workflows/ci.yml#L42 the CI will error on it instead of warn.

huynhdev24 (approved):

merceyz (approved):

PR #432, opened 2024-03-17
feat: verify integrity signature when downloading from npm registry

aduh95:
When the user has not provided any hash (so when running `corepack up`/`corepack use …`), and the package manager is downloaded from the npm registry, we can verify the signature.

Related to https://github.com/nodejs/corepack/issues/10.

BREAKING CHANGE: attempting to download a version from the npm registry (or a mirror) that was published using the now deprecated PGP signature (Yarn ≤1.22.18, Yarn Berry ≤4.0.0-rc.43, pnpm ≤8.4.0) without providing a hash will trigger an error. Users can disable the signature verification by setting `COREPACK_INTEGRITY_KEYS=""`.


config.json
```diff
@@ -160,5 +160,16 @@
...
+        "expires": null,
+        "keyid": "SHA256:jl3bwswu80PjjokCgh0o2w5c2U4LhQAE57gj9cz1kzA",
+        "keytype": "ecdsa-sha2-nistp256",
+        "scheme": "ecdsa-sha2-nistp256",
+        "key": "MFkwEwYHKoZIzj0CAQYIKoZIzj0DAQcDQgAE1Olb3zMAFFxXKHiIkQO5cJ3Yhl5i6UPp+IhuteBJbuHcA5UogKo0EWtlWwW6KSaKoTNEYL7JlCQiVnkhBktUgg=="
+      }
+    ]
   }
-}
+}
```
aduh95, line 175:
```suggestion
}

```

sources/types.ts
```diff
@@ -96,6 +96,10 @@ export interface Config {
       };
     };
   };
+
+  keys: {
+    [registry: string]: Array<{"expires": null, "keyid": string, "keytype": string, "scheme": string, "key": string}>;
+  };
```
arcanis, line 102: nit:
```suggestion
  keys: {
    [registry: string]: Array<{
      expires: null;
      keyid: string;
      keytype: string;
      scheme: string;
      key: string;
    }>;
  };
```

sources/corepackUtils.ts
```diff
@@ -200,7 +204,7 @@ export async function installVersion(installTarget: string, locator: Locator, {s
 
   stream.pipe(sendTo);
 
-  const algo = build[0] ?? `sha256`;
+  const algo = build[0] ?? `sha512`;
```
arcanis, line 207: Why switching to sha512?

aduh95: Because that’s what npm registry for the signature: they sign a combination of the package name & version and the SHA512 of the package archive.

sources/corepackUtils.ts
```diff
@@ -234,7 +238,37 @@ export async function installVersion(installTarget: string, locator: Locator, {s
     }
   }
 
-  const actualHash = hash.digest(`hex`);
+  const actualHash: string = hash.digest(`hex`);
+  if (!build[1]) {
+    const registry = getRegistryFromPackageManagerSpec(spec);
+    if (registry.type === `npm` && !process.env.COREPACK_NPM_REGISTRY) {
```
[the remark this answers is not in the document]
aduh95: What would you put inside? If it’s a JSON object we could let user defined several keys for several registries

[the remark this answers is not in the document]
aduh95: I've changed that to be `COREPACK_INTEGRITY_KEYS` which takes a JSON string in case users want/need to override the internal config.

sources/corepackUtils.ts
```diff
@@ -8,6 +8,7 @@ import path                                                    from 'path';
 import semver                                                  from 'semver';
 import {setTimeout as setTimeoutPromise}                       from 'timers/promises';
 
+
```
aduh95, line 11:
```suggestion
```

arcanis (approved):

tests/main.test.ts
```diff
@@ -856,3 +857,171 @@ for (const authType of [`COREPACK_NPM_REGISTRY`, `COREPACK_NPM_TOKEN`, `COREPACK
...
+        stdout: `Installing yarn@1.9998.9999 in the project...\n\nyarn: Hello from custom registry\n`,
+        stderr: ``,
+      });
+      await expect(runCli(cwd, [`use`, `yarn@latest`], true)).resolves.toMatchObject({
+        exitCode: 0,
+        stdout: `Installing yarn@5.9999.9999 in the project...\n\nyarn: Hello from custom registry\n`,
+        stderr: ``,
+      });
+    });
+  });
```
aduh95, line 906:
```suggestion
  }, 180_000);
```

PR #439, opened 2024-03-25
feat: download only `yarn.js` from npm

### LINKED ISSUE #435: Hash validation failed for yarn when `COREPACK_NPM_REGISTRY` is set on one side
Currently, yarn berry could be downloaded as two form, tar or js, depending on whether `COREPACK_NPM_REGISTRY` is set. However, the hash validation step simply calculates the shasum of download stream, and obviously the hashes mismatch.

For example, shasums for yarn 4.1.1 are:
```
61b9f63c5edc625867eeda36190a4efebdf7840052db5f6583e301a9d228eb43  cli-dist-4.1.1.tgz
f3cc0eda8e5560e529c7147565b30faa43b4e472d90e8634d7134a37c7f59781  yarn.js
```

Reproduce steps:
1. Call `corepack use yarn@4.1.1` without `COREPACK_NPM_REGISTRY` being set. Hash will be written to `package.json`
```
"packageManager": "yarn@4.1.1+sha256.f3cc0eda8e5560e529c7147565b30faa43b4e472d90e8634d7134a37c7f59781"
```
2. Copy the project to an environment where `COREPACK_NPM_REGISTRY` is set.
3. Call `corepack yarn`, following error will be thrown:

```
Internal Error: Mismatch hashes. Expected f3cc0eda8e5560e529c7147565b30faa43b4e472d90e8634d7134a37c7f59781, got 61b9f63c5edc625867eeda36190a4efebdf7840052db5f6583e301a9d228eb43
```

zhyupe:
Added `bin` property to `NpmRegistrySpec`. When installing yarn berry from custom registry, it'll only install the bin file and calculate hash of the bin file itself.

Fix #435

### DIFF
--- a/sources/corepackUtils.ts
+++ b/sources/corepackUtils.ts
@@ -155,10 +155,7 @@ async function download(installTarget: string, url: string, algo: string, binPat
       cwd: tmpFolder,
       filter: binPath ? path => {
         const pos = path.indexOf(`/`);
-        if (pos === -1 || path.slice(pos + 1) !== binPath)
-          return false;
-
-        return true;
+        return pos !== -1 && path.slice(pos + 1) === binPath;
       } : undefined,
     });
   } else if (ext === `.js`) {
--- a/sources/corepackUtils.ts
+++ b/sources/corepackUtils.ts
@@ -161,12 +161,9 @@ async function download(installTarget: string, url: string, algo: string, binPat
   await once(sendTo, `finish`);
 
   if (binPath) {
-    await once(sendTo, `finish`);
-
     const downloadedBin = path.join(downloadFolder, binPath);
     if (!fs.existsSync(downloadedBin))
       throw new Error(`Cannot locate '${binPath}' in downloaded tarball`);
-
 
     // Create a new folder which only contains the bin file
     const tmpFolder = folderUtils.getTemporaryFolder(installTarget);
@@ -181,7 +178,7 @@ async function download(installTarget: string, url: string, algo: string, binPat
     // Calculate the hash of the bin file
     const fileStream = fs.createReadStream(outputFile);
     const fileHash = fileStream.pipe(createHash(algo));
-    await once(fileStream, `finish`);
+    await once(fileStream, `close`);
 
     return {
       tmpFolder,
--- a/sources/corepackUtils.ts
+++ b/sources/corepackUtils.ts
@@ -172,11 +172,15 @@ async function download(installTarget: string, url: string, algo: string, binPat
 
   if (binPath) {
     const downloadedBin = path.join(tmpFolder, binPath);
-    if (!fs.existsSync(downloadedBin))
-      throw new Error(`Cannot locate '${binPath}' in downloaded tarball`);
-
     outputFile = path.join(tmpFolder, path.basename(downloadedBin));
-    await rename(downloadedBin, outputFile);
+    try {
+      await rename(downloadedBin, outputFile);
+    } catch (err) {
+      if ((err as nodeUtils.NodeError)?.code === `ENOENT`) {
+        throw new Error(`Cannot locate '${binPath}' in downloaded tarball`, { cause: err });
+      }
+      throw err;
+    }
 
     // Calculate the hash of the bin file
     const fileStream = fs.createReadStream(outputFile);
--- a/sources/corepackUtils.ts
+++ b/sources/corepackUtils.ts
@@ -191,7 +191,7 @@ async function download(installTarget: string, url: string, algo: string, binPat
   return {
     tmpFolder,
     outputFile,
-    hash: hash.digest(`hex`),
+    hash: hash!.digest(`hex`),
   };
 }
 
--- a/sources/corepackUtils.ts
+++ b/sources/corepackUtils.ts
@@ -1,21 +1,21 @@
-import {createHash}                                            from 'crypto';
-import {once}                                                  from 'events';
-import {FileHandle}                                            from 'fs/promises';
-import fs                                                      from 'fs';
-import type {Dir}                                              from 'fs';
-import Module                                                  from 'module';
-import path                                                    from 'path';
-import semver                                                  from 'semver';
-import {setTimeout as setTimeoutPromise}                       from 'timers/promises';
+import {createHash}                                                          from 'crypto';
+import {once}                                                                from 'events';
+import {FileHandle}                                                          from 'fs/promises';
+import fs                                                                    from 'fs';
+import type {Dir}                                                            from 'fs';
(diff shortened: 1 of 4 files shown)


aduh95:
Have you considered using `filter` option of [`tar.x`](https://github.com/isaacs/node-tar?tab=readme-ov-file#tarxoptions-filelist-callback-alias-tarextract)?

sources/corepackUtils.ts
```diff
@@ -132,6 +132,71 @@ function isValidBinSpec(x: unknown): x is BinSpec {
...
+  const streamHash = stream.pipe(createHash(algo));
+  await once(sendTo, `finish`);
+
+  if (binPath) {
+    const downloadedBin = path.join(tmpFolder, binPath);
+    if (!fs.existsSync(downloadedBin))
+      throw new Error(`Cannot locate '${binPath}' in downloaded tarball`);
+
+    outputFile = path.join(tmpFolder, path.basename(downloadedBin));
+    await rename(downloadedBin, outputFile);
```
aduh95, line 179:
```suggestion
    outputFile = path.join(tmpFolder, path.basename(downloadedBin));
    try {
      if (process.platform === `win32`) {
        await renameUnderWindows(tmpFolder, installFolder);
      } else {
        await fs.promises.rename(tmpFolder, installFolder);
      }
    } catch (err) {
      if ((err as nodeUtils.NodeError)?.code === `ENOENT`) {
        throw new Error(`Cannot locate '${binPath}' in downloaded tarball`, { cause: err });
      }
      throw err;
    }
```

sources/corepackUtils.ts
```diff
@@ -132,6 +132,75 @@ function isValidBinSpec(x: unknown): x is BinSpec {
...
+      outputFile,
+      hash: fileHash.digest(`hex`),
+    };
+  }
+
+  return {
+    tmpFolder,
+    outputFile,
+    hash: streamHash.digest(`hex`),
+  };
```
aduh95, line 201: Given those return are almost exactly similar, it might make more sense to use only one, wdyt?

zhyupe: okay

sources/corepackUtils.ts
```diff
@@ -132,6 +132,74 @@ function isValidBinSpec(x: unknown): x is BinSpec {
...
+        return true;
+      } : undefined,
+    });
+  } else if (ext === `.js`) {
+    outputFile = path.join(tmpFolder, path.posix.basename(parsedUrl.pathname));
+    sendTo = fs.createWriteStream(outputFile);
+  }
+  stream.pipe(sendTo);
+
+  const streamHash = stream.pipe(createHash(algo));
```
arcanis, line 170: We don't need `streamHash` if `binPath` is set, right? So we can skip checksuming it:
```suggestion
  const streamHash = !binPath ? stream.pipe(createHash(algo)) : null;
```

sources/corepackUtils.ts
```diff
@@ -132,6 +132,74 @@ function isValidBinSpec(x: unknown): x is BinSpec {
...
+  } else if (ext === `.js`) {
+    outputFile = path.join(tmpFolder, path.posix.basename(parsedUrl.pathname));
+    sendTo = fs.createWriteStream(outputFile);
+  }
+  stream.pipe(sendTo);
+
+  const streamHash = !binPath ? stream.pipe(createHash(algo)) : null;
+  await once(sendTo, `finish`);
+
+  let hash: string;
```
aduh95, line 173: nit
```suggestion
  let hash = !binPath ? stream.pipe(createHash(algo)) : null;
  await once(sendTo, `finish`);

```

sources/corepackUtils.ts
```diff
@@ -132,6 +132,74 @@ function isValidBinSpec(x: unknown): x is BinSpec {
...
+
+    hash = fileHash.digest(`hex`);
+  } else {
+    hash = streamHash.digest(`hex`);
+  }
+
+  return {
+    tmpFolder,
+    outputFile,
+    hash,
```
aduh95, line 199:
```suggestion
    hash = fileStream.pipe(createHash(algo));
    await once(fileStream, `close`);
  }

  return {
    tmpFolder,
    outputFile,
    hash: hash.digest(`hex`),
```

sources/corepackUtils.ts
```diff
@@ -132,6 +132,69 @@ function isValidBinSpec(x: unknown): x is BinSpec {
...
+    // Calculate the hash of the bin file
+    const fileStream = fs.createReadStream(outputFile);
+    hash = fileStream.pipe(createHash(algo));
+    await once(fileStream, `close`);
+  }
+
+  return {
+    tmpFolder,
+    outputFile,
+    hash: hash.digest(`hex`),
```
aduh95, line 194:
```suggestion
    hash: hash!.digest(`hex`),
```

sources/corepackUtils.ts
```diff
@@ -132,6 +132,69 @@ function isValidBinSpec(x: unknown): x is BinSpec {
...
+    const {default: tar} = await import(`tar`);
+    sendTo = tar.x({
+      strip: 1,
+      cwd: tmpFolder,
+      filter: binPath ? path => {
+        const pos = path.indexOf(`/`);
+        if (pos === -1 || path.slice(pos + 1) !== binPath)
+          return false;
+
+        return true;
```
aduh95, line 161:
```suggestion
        return pos !== -1 && path.slice(pos + 1) === binPath;
```

aduh95:
Other than tests, is there something else to be done before this can be marked as ready for reviews?

tests/main.test.ts
```diff
@@ -787,19 +787,19 @@ it(`should download yarn berry from custom registry`, async () => {
     process.env.COREPACK_ENABLE_DOWNLOAD_PROMPT = `1`;
 
     await xfs.writeJsonPromise(ppath.join(cwd, `package.json` as Filename), {
-      packageManager: `yarn@3.0.0`,
+      packageManager: `yarn@4.1.1+sha256.f3cc0eda8e5560e529c7147565b30faa43b4e472d90e8634d7134a37c7f59781`,
```
aduh95, line 790: Let's reuse the same version as in https://github.com/nodejs/corepack/blob/8b6c6d4b2b7a9d61ae6c33c07e12354bd5afc2ba/tests/main.test.ts#L121
```suggestion
      packageManager: `yarn@3.0.0-rc.2+sha224.f83f6d1cbfac10ba6b516a62ccd2a72ccd857aa6c514d1cd7185ec60`,
```

zhyupe:
@aduh95 Waiting for #445 being merged or the tests would fail

aduh95 (approved):

PR #441, opened 2024-03-26
chore: use `pipeline` from `node:stream/promises`

aduh95:
Because Corepack predates the Stream Promise API, the code was using `once(stream, 'finish')`, which is arguably not as self-explanatory as `pipeline()` API.


arcanis (approved):

aduh95:
There are a few tests falling, and it looks like this is going to conflict with https://github.com/nodejs/corepack/pull/439. Giving up for now.

Issue #422, opened 2024-03-10
Don't change file if not necessary


cdtut:
`corepack install --global` always writes to file. Don't touch file if not necessary including if there is not change to a version. And output should be different and say `yarn@version already installed` not `Installing yarn@version...`.

aduh95:
Can you clarify what you mean by that? Any chance you could include a list of commands that would show the problem?

cdtut:
`corepack install --global yarn` adds yarn to `~/.cache/node/corepack/lastKnownGood.json`. Do it again and same file is written even if no change.

And output still says `Installing` but was already installed.

Issue #183, opened 2022-09-20
Corepack should have an option to be able to run on a readonly docker container


marcmillien:
## Description

I currently have the following `Dockerfile`:
```
FROM node:18-bullseye              
RUN corepack enable yarn
```

My production environment runs this image in readonly, this way:
```
docker run --rm -it --read-only --mount 'type=tmpfs,dst=/tmp' IMAGE_BUILT bash
```

The problem is we can't instruct `corepack` to not write cache folder leading to the following error running yarn:
```
root@5cd6509d006b:/# yarn --version
Internal Error: ENOENT: no such file or directory, mkdir '/root/.cache'
Error: ENOENT: no such file or directory, mkdir '/root/.cache'
```

My current workaround is to use `/tmp` (monkey patch, this folder is not the best from security perspective):
```
FROM node:18-bullseye              
ENV COREPACK_HOME /tmp/corepack
RUN corepack enable yarn
```

I think, the good solution from a security perspective would be to have an option to instruct corepack not to write anything.

aduh95:
> My current workaround is to use `/tmp` (monkey patch, this folder is not the best from security perspective)

Can you explain further? What's the security issue with `/tmp`?

> I think, the good solution from a security perspective would be to have an option to instruct corepack not to write anything.

I'm not sure that's even possible, the downloaded code needs to be somewhere for Node.js to execute it – or we would need to ship a whole VFS implementation which I'm not sure is worth it. Yarn being only one file, that would not be necessary, and we would "just" need to monkey patch the CJS loader, however for the other package managers that ship in several files, that would get complicated I think.

merceyz:
While it isn't what you're asking for, you could populate the cache before it becomes readonly
```Dockerfile
FROM node:18-bullseye
RUN corepack enable yarn && yarn -v
```

PR #446, opened 2024-04-01
Separate read and write operations on lastKnownGood.json

das7pad:
#### Description

Closes https://github.com/nodejs/corepack/issues/183
Closes https://github.com/nodejs/corepack/issues/422

This PR is separating read and write operations on `lastKnownGood.json`. It will also skip overwriting `lastKnownGood.json` with the same content.

#### Review

This PR is best reviewed with white space changes ignored due to indentation changes from removing try/catch/finally blocks.

I refactored the read and write operations into self contained functions. The re-use of a `r+` file handle does not work anymore when separating the two kinds of operations.

The content of `lastKnownGood.json` is validated/sanitized when reading. Any consumers can operate on the `Record<string, string>` object directly. This greatly simplified the logic in `activatePackageManager` and `getLastKnownGoodFromFileContent`.

I could not get the "per-project" install to work fully offline/read-only in a straightforward way (just running `corepack install`). Running `corepack install` only caches the requested binary from `package.json -> packageManager`. When invoking the package-manager (yarn) via `corepack yarn`, corepack tries to discover the latest version to prepare a `fallbackLocator` in `Engine.executePackageManagerRequest` first, but without a `lastKnownGood.json` (`corepack install` does not prepare it), it attempts a network request. The "code fix" here is to defer the `fallbackLocator` work until after discovering any `package.json` file. This is a major undertaking and something for another day. I left a note for this in a comment in the "per-project" test, next to the workaround: running `corepack yarn --version` once populates the `lastKnownGood.json` cache.

### DIFF
--- a/sources/Engine.ts
+++ b/sources/Engine.ts
@@ -1,5 +1,4 @@
 import {UsageError}                                           from 'clipanion';
-import type {FileHandle}                                      from 'fs/promises';
 import fs                                                     from 'fs';
 import path                                                   from 'path';
 import process                                                from 'process';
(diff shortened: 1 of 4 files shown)


sources/Engine.ts
```diff
@@ -150,54 +157,32 @@ export class Engine {
...
+      // If for some reason, we cannot update the last known good file, we
+      // ignore the error.
     }
+
+    return reference;
+  }
+
+  async activatePackageManager(locator: Locator) {
+    const lastKnownGood = await getLastKnownGood();
+    await activatePackageManager(lastKnownGood, locator);
```
aduh95, line 185: This pattern of reading a file, closing it, then writing to overwrite its content can result in race conditions if there are two processes trying to do that operation in parallel. That's why I refactored the code to use `FileHandle`s in https://github.com/nodejs/corepack/pull/364. That being said, maybe supporting ro FS is more important than supporting running two `corepack` processes in parallel.

There's a failing test ("should support running globally") to address.

tests/main.test.ts
```diff
@@ -476,6 +476,96 @@ it(`should support disabling the network accesses from the environment`, async (
...
+        stdout: `2.2.2\n`,
+        stderr: ``,
+        exitCode: 0,
+      });
+      // $ yarn --version
+      await expect(spawnCmd(installDir, `yarn`, [`--version`])).resolves.toMatchObject({
+        stdout: `2.2.2\n`,
+        stderr: ``,
+        exitCode: 0,
+      });
```
aduh95, line 564: That's not supported on Windows (because Windows doesn't use the PATH variable same as other OSes I think), let's remove it

aduh95:
> corepack tries to discover the latest version to prepare a `fallbackLocator` in `Engine.executePackageManagerRequest` first, but without a `lastKnownGood.json` (`corepack install` does not prepare it), it attempts a network request

I wonder if https://github.com/nodejs/corepack/pull/432/files#diff-ac2e60980d41d74acab8d4ec9d41b6bd52b08f056b3ea65f89952de64eca3601 did not fix that already.

aduh95 (approved):

Issue #417, opened 2024-03-06
Using corepack behind corporate proxy and injected certificate


adurand-ctie:
Hi,
I'm willing to use API Platform (https://api-platform.com/) and the Dockerfile of the pwa component requires corepack:
```
RUN corepack enable && \
	corepack prepare --activate pnpm@latest && \
	pnpm config -g set store-dir /.pnpm-store
```
The base image is node:20-alpine.

I'm behind a corporate proxy, so I've added the following before using corepack:
```
RUN set HTTP_PROXY=http://proxy.etat.lu:8080
RUN set HTTPS_PROXY=http://proxy.etat.lu:8080
RUN set ALL_PROXY=http://proxy.etat.lu:8080
```
We also have a "traffic interceptor" that watches the whole traffic, so I added its certificate (which is already injected in the image):
`RUN set NODE_EXTRA_CA_CERTS=/usr/local/share/ca-certificates/etat.crt`

However, I'm still running into issues and receive the error below upon executing `corepack prepare --activate pnpm@latest`:
```
Internal Error: Error when performing the request to https://registry.npmjs.org/pnpm; for troubleshooting help, see https://github.com/nodejs/corepack#troubleshooting
    at ClientRequest.<anonymous> (/usr/local/lib/node_modules/corepack/dist/lib/corepack.cjs:42195:14)
    at ClientRequest.emit (node:events:518:28)
    at TLSSocket.socketErrorListener (node:_http_client:495:9)
    at TLSSocket.emit (node:events:518:28)
    at emitErrorNT (node:internal/streams/destroy:169:8)
    at emitErrorCloseNT (node:internal/streams/destroy:128:3)
    at process.processTicksAndRejections (node:internal/process/task_queues:82:21)
The command '/bin/sh -c corepack prepare --activate pnpm@latest' returned a non-zero code: 1
ERROR: Service 'pwa' failed to build : Build failed
```
Anybody could help me figure out what I'm missing here?

Thanks in advance.
Adrien.

merceyz:
Probably related to https://github.com/nodejs/undici/issues/2200.

adurand-ctie:
> Probably related to [nodejs/undici#2200](https://github.com/nodejs/undici/issues/2200).

To overcome this, should I try `RUN set NODE_TLS_REJECT_UNAUTHORIZED=0`?
If this is the case, it didn't work :(

aduh95:
[...]

PR #451, opened 2024-04-06
fix: correctly set `Dispatcher` prototype for `ProxyAgent`

aduh95:
In an attempt to bundle only a subset of Undici code, we forgot to take some side-effect.

Fixes: https://github.com/nodejs/corepack/issues/417
Closes: https://github.com/nodejs/corepack/pull/449

### DIFF
--- a/sources/httpUtils.ts
+++ b/sources/httpUtils.ts
@@ -92,6 +92,8 @@ export async function fetchUrlStream(input: string | URL, init?: RequestInit) {
   return stream;
 }
 
+let ProxyAgent: typeof import('undici').ProxyAgent;
+
 async function getProxyAgent(input: string | URL) {
   const {getProxyForUrl} = await import(`proxy-from-env`);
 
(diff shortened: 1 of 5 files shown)


sources/httpUtils.ts
```diff
@@ -100,11 +102,20 @@ async function getProxyAgent(input: string | URL) {
...
+  if (ProxyAgent == null) {
+    // Doing a deep import here since undici isn't tree-shakeable
+    const [api, Dispatcher, _ProxyAgent] = await Promise.all([
+      // @ts-expect-error internal module is untyped
+      import(`undici/lib/api/index.js`),
+      // @ts-expect-error internal module is untyped
+      import(`undici/lib/dispatcher/dispatcher.js`),
+      // @ts-expect-error internal module is untyped
+      import(`undici/lib/dispatcher/proxy-agent.js`),
+    ]);
```
arcanis, line 114: Are we doing all that just to reclaim some bytes? It seems a little dangerous 🤔

merceyz: 304000 of them but yes.

sources/httpUtils.ts
```diff
@@ -100,11 +102,20 @@ async function getProxyAgent(input: string | URL) {
...
   if (!proxy) return undefined;
 
-  // Doing a deep import here since undici isn't tree-shakeable
-  const {default: ProxyAgent} = (await import(
-    // @ts-expect-error No types for this specific file
-    `undici/lib/proxy-agent.js`
-  )) as { default: typeof import('undici').ProxyAgent };
+  if (ProxyAgent == null) {
+    // Doing a deep import here since undici isn't tree-shakeable
+    const [api, Dispatcher, _ProxyAgent] = await Promise.all([
```
onigoetz, line 107: Do these need to be loaded in a particular order?

I would recommend putting them in a separate file and just exposing the ProxyAgent from that file and add a comment on why those files are needed.

aduh95: No, any order, that's why I'm loading them concurrently.

sources/httpUtils.ts
aduh95: I've added tests, hopefully those ensure we'll catch bugs if this hack breaks again in the future.

arcanis (approved):

merceyz (approved):

PR #454, opened 2024-04-11
fix: Incorrect authorization prefix for basic auth, and undocumented env var

plumpNation:
Changes auth type header strings to use `Basic` prefix for base64 encoded username:password.

In addition, it seems like an undocumented env var `COREPACK_NPM_USER` is being used in tests, meaning that this value will be `undefined` and the user will always get a 401 as the documented env var is `COREPACK_NPM_USERNAME`.
This bit me when I was using a test to help debug why my COREPACK_NPM_REGISTRY was being ignored when running `yarn --version` in real life.

### DIFF
--- a/sources/httpUtils.ts
+++ b/sources/httpUtils.ts
@@ -29,12 +29,7 @@ async function fetch(input: string | URL, init?: RequestInit) {
     input.username = input.password = ``;
   }
 
-  const registries = [
-    process.env.COREPACK_NPM_REGISTRY,
-    DEFAULT_NPM_REGISTRY_URL,
-  ].filter(Boolean);
-
-  if (registries.includes(input.origin) && process.env.COREPACK_NPM_TOKEN) {
+  if (input.origin === (process.env.COREPACK_NPM_REGISTRY || DEFAULT_NPM_REGISTRY_URL) && process.env.COREPACK_NPM_TOKEN) {
     headers =  {
       ...headers,
       authorization: `Bearer ${process.env.COREPACK_NPM_TOKEN}`,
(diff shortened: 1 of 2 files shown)


sources/httpUtils.ts
```diff
@@ -16,28 +16,28 @@ async function fetch(input: string | URL, init?: RequestInit) {
...
       ...headers,
-      authorization: `Bearer ${Buffer.from(`${username}:${password}`).toString(`base64`)}`,
+      authorization: `Basic ${Buffer.from(`${username}:${password}`).toString(`base64`)}`,
     };
+
     input.username = input.password = ``;
-  } else if (input.origin === process.env.COREPACK_NPM_REGISTRY || DEFAULT_NPM_REGISTRY_URL) {
+  }
+
+  if (input.origin === process.env.COREPACK_NPM_REGISTRY || DEFAULT_NPM_REGISTRY_URL) {
```
aduh95, line 32: I realize this is not related to this PR, but it looks like the condition would be always true as it is written. We should fix it.
```suggestion
  if (input.origin === process.env.COREPACK_NPM_REGISTRY || input.origin === DEFAULT_NPM_REGISTRY_URL) {
```

plumpNation: I will look at this. Thanks for the feedback.

sources/httpUtils.ts
```diff
@@ -16,27 +16,30 @@ async function fetch(input: string | URL, init?: RequestInit) {
...
-      };
-    }
   }
 
+  const registries = [
+    process.env.COREPACK_NPM_REGISTRY,
+    DEFAULT_NPM_REGISTRY_URL,
+  ].filter(Boolean);
+
+  if (registries.includes(input.origin) && process.env.COREPACK_NPM_TOKEN) {
```
aduh95, line 37:
```suggestion
  if (input.origin === (process.env.COREPACK_NPM_REGISTRY || DEFAULT_NPM_REGISTRY_URL) && process.env.COREPACK_NPM_TOKEN) {
```

PR #456, opened 2024-04-13
fix: add path to `package.json` in error message

aduh95:
I had to help someone who couldn't understand why they were getting this error message on an empty directory; turned out they had a `package.json` in their home directory (no idea how it got there) that defined a `packageManager` field. I think adding some explanation would help.

### DIFF
--- a/sources/Engine.ts
+++ b/sources/Engine.ts
@@ -278,7 +278,7 @@ export class Engine {
             if (transparent) {
               return fallbackDescriptor;
             } else {
-              throw new UsageError(`This project is configured to use ${result.spec.name}`);
+              throw new UsageError(`This project is configured to use ${result.spec.name} because ${result.target} defines a "packageManager" field`);
             }
           } else {
             return result.spec;
--- a/sources/Engine.ts
+++ b/sources/Engine.ts
@@ -278,7 +278,7 @@ export class Engine {
             if (transparent) {
               return fallbackDescriptor;
             } else {
-              throw new UsageError(`This project is configured to use ${result.spec.name} because ${result.target} defines a "packageManager" field`);
+              throw new UsageError(`This project is configured to use ${result.spec.name} because ${result.target} has a "packageManager" field`);
             }
           } else {
             return result.spec;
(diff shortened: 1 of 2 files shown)


arcanis (approved):

arcanis:
> no idea how it got there

That's an issue we sometimes have with `yarnPath` as well. Perhaps we should make `corepack use` fail if run in the home directory, unless a special CLI flag is set (like `rm --no-preserve-root`)?

sources/Engine.ts
```diff
@@ -278,7 +278,7 @@ export class Engine {
             if (transparent) {
               return fallbackDescriptor;
             } else {
-              throw new UsageError(`This project is configured to use ${result.spec.name}`);
+              throw new UsageError(`This project is configured to use ${result.spec.name} because ${result.target} defines a "packageManager" field`);
```
styfle, line 281:
```suggestion
              throw new UsageError(`This project is configured to use ${result.spec.name} because ${result.target} has a "packageManager" field`);
```

defines => has

I think this reads a little better

tests/main.test.ts
```diff
@@ -419,7 +419,9 @@ it(`should refuse to run a different package manager within a configured project
     process.env.FORCE_COLOR = `0`;
 
     await expect(runCli(cwd, [`pnpm`, `--version`])).resolves.toMatchObject({
-      stdout: `Usage Error: This project is configured to use yarn\n\n$ pnpm ...\n`,
+      stdout: `Usage Error: This project is configured to use yarn because ${
+        ppath.join(cwd, `package.json` as Filename)
+      } defines a "packageManager" field\n\n$ pnpm ...\n`,
```
styfle, line 424:
```suggestion
      } has a "packageManager" field\n\n$ pnpm ...\n`,
```

styfle (approved):

merceyz (approved):

tests/main.test.ts
```diff
@@ -419,7 +420,10 @@ it(`should refuse to run a different package manager within a configured project
     process.env.FORCE_COLOR = `0`;
 
     await expect(runCli(cwd, [`pnpm`, `--version`])).resolves.toMatchObject({
-      stdout: `Usage Error: This project is configured to use yarn\n\n$ pnpm ...\n`,
+      stdout: `Usage Error: This project is configured to use yarn because ${
+        // ppath and xfs do not format Windows correctly, the regex fixes that:
```
merceyz, line 424: You would use `npath.fromPortablePath` to handle that.

Issue #458, opened 2024-04-15
Frequent fetch() errors


xconverge:
I am encountering this a LOT intermittently during CI for my projects. I can rerun CI and then it is all fine. I have looked at the troubleshooting referenced but I am not behind a proxy so am not really sure what else it could be. My suspicion is some sort of load oriented timeout. It could be intermittent internet too, but without seeing the reason for the exception it is hard for me to debug

```
Internal Error: Error when performing the request to https://registry.npmjs.org/yarn; for troubleshooting help, see https://github.com/nodejs/corepack#troubleshooting
    at fetch (/usr/local/lib/node_modules/corepack/dist/lib/corepack.cjs:22882:11)
    at process.processTicksAndRejections (node:internal/process/task_queues:95:5)
    at async fetchAsJson (/usr/local/lib/node_modules/corepack/dist/lib/corepack.cjs:22896:20)
    at async fetchLatestStableVersion (/usr/local/lib/node_modules/corepack/dist/lib/corepack.cjs:22948:20)
    at async fetchLatestStableVersion2 (/usr/local/lib/node_modules/corepack/dist/lib/corepack.cjs:22971:14)
    at async Engine.getDefaultVersion (/usr/local/lib/node_modules/corepack/dist/lib/corepack.cjs:23349:25)
    at async executePackageManagerRequest (/usr/local/lib/node_modules/corepack/dist/lib/corepack.cjs:24207:28)
    at async BinaryCommand.validateAndExecute (/usr/local/lib/node_modules/corepack/dist/lib/corepack.cjs:21173:22)
    at async _Cli.run (/usr/local/lib/node_modules/corepack/dist/lib/corepack.cjs:22148:18)
    at async Object.runMain (/usr/local/lib/node_modules/corepack/dist/lib/corepack.cjs:24279:12)
```

I see there is some data in the throw here: https://github.com/nodejs/corepack/blob/142319056424b1e0da2bdbe801c52c5910023707/sources/httpUtils.ts#L48

Is there a way for me to see the contents of this easily, I am not sure where it ends up. If not (I doubt this will be the answer) could we perhaps move the err to the text instead?

xconverge:
So far I have tried:

1. Changing from ISP provided DNS server to cloudflare or google DNS on the server
2. `export NODE_OPTIONS="--dns-result-order=ipv4first --no-network-family-autoselection"`

With no change in behavior

Not sure I have any other knobs to turn on my end...

[...]

PR #463, opened 2024-04-20
fix: add error cause to error output

aduh95:
Refs: https://github.com/nodejs/corepack/issues/458


aduh95:
Better solved by https://github.com/nodejs/corepack/pull/430.

PR #466, opened 2024-04-23
fix: COREPACK_NPM_REGISTRY should allow for username/password auth

Keysox:
I noticed when using this package, if you set `COREPACK_NPM_REGISTRY`, authentication wasn't working as expected.

After debugging and reading through the code, it appears that when the following code was run:
```
if (typeof input === `string`)
    input = new URL(input);
```
`input.username` would become an empty string. This caused `process.env.COREPACK_NPM_USERNAME` to not be used. 

By switching `??` to `||`, the authorization header should now be sent to the registry as expected!

### DIFF
--- a/sources/httpUtils.ts
+++ b/sources/httpUtils.ts
@@ -17,8 +17,8 @@ async function fetch(input: string | URL, init?: RequestInit) {
 
   let headers = init?.headers;
 
-  const username: string | undefined = input.username ?? process.env.COREPACK_NPM_USERNAME;
-  const password: string | undefined = input.password ?? process.env.COREPACK_NPM_PASSWORD;
+  const username: string | undefined = input.username || process.env.COREPACK_NPM_USERNAME;
+  const password: string | undefined = input.password || process.env.COREPACK_NPM_PASSWORD;
 
   if (username || password) {
     headers =  {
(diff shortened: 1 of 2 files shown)


aduh95:
Thanks for the PR, that makes sense. Do you know why the tests do not catch that? We're checking against a custom mock repository, which should validate authentication works 🤔

aduh95:
Well I found the problem: the mock registry would skip auth validation if none was provided, Windows 95 style 🤦‍♂️ Can you apply the following diff to ensure we don't regress:

```diff
diff --git a/tests/_registryServer.mjs b/tests/_registryServer.mjs
index ff0be26..08fa8f4 100644
--- a/tests/_registryServer.mjs
+++ b/tests/_registryServer.mjs
@@ -116,8 +116,9 @@ const server = createServer((req, res) => {
   const auth = req.headers.authorization;
 
   if (
-    (auth?.startsWith(`Bearer `) && auth.slice(`Bearer `.length) !== TOKEN_MOCK) ||
-    (auth?.startsWith(`Basic `) && Buffer.from(auth.slice(`Basic `.length), `base64`).toString() !== `user:pass`)
+    auth == null ||
+    (auth.startsWith(`Bearer `) && auth.slice(`Bearer `.length) !== TOKEN_MOCK) ||
+    (auth.startsWith(`Basic `) && Buffer.from(auth.slice(`Basic `.length), `base64`).toString() !== `user:pass`)
   ) {
     res.writeHead(401).end(`Unauthorized`);
     return;
```

PR #491, opened 2024-06-06
chore(ci): better check for Nocks changes

aduh95:
When there are no changes, asking Git to compare the db file directly will always report change even though the actual data stored is the same. Instead, we can compare a SQL dump to detect actual changes.

### DIFF
--- a/.github/workflows/update-nock-files.yml
+++ b/.github/workflows/update-nock-files.yml
@@ -50,13 +50,11 @@ jobs:
       - name: Check if anything has changed
         id: contains-changes
         run: |
-          sqlite3 tests/nocks.db .dump > tests/nocks.sql
-          git add tests/nocks.sql
+          sqlite3 tests/nocks.db .dump > /tmp/before.sql
           cp tests/nocks.db tests/nocks.db.new
           git checkout HEAD -- tests/nocks.db
-          sqlite3 tests/nocks.db .dump > tests/nocks.sql
-          echo "result=$(git --no-pager diff --quiet -- tests/nocks.sql || echo "yes")" >> "$GITHUB_OUTPUT"
-          git rm -f tests/nocks.sql
+          sqlite3 tests/nocks.db .dump > /tmp/after.sql
+          echo "result=$(git --no-pager diff --quiet --no-index /tmp/before.sql /tmp/after.sql || echo "yes")" >> "$GITHUB_OUTPUT"
           mv tests/nocks.db.new tests/nocks.db
         shell: bash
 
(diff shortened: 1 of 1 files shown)


.github/workflows/update-nock-files.yml
```diff
@@ -49,7 +49,15 @@ jobs:
 
       - name: Check if anything has changed
         id: contains-changes
-        run: echo "result=$(git --no-pager diff --quiet -- tests/nocks.db || echo "yes")" >> $GITHUB_OUTPUT
+        run: |
+          sqlite3 tests/nocks.db .dump > tests/nocks.sql
+          git add tests/nocks.sql
+          cp tests/nocks.db tests/nocks.db.new
+          git checkout HEAD -- tests/nocks.db
```
arcanis, line 56: nit: Rather than the `cp`, you could use `git show` to dump the old content in a `/tmp` file without mutating the real `nocks.db`.

arcanis (approved):

.github/workflows/update-nock-files.yml
aduh95: I've considered it, and as I was not sure what would be the behavior of `git show` with binary files, I decided to do that `cp` dance instead. I don't feel strongly either way, happy to change it if you think it's better

arcanis: I'd expect it to be possible but I can't find the right syntax locally, so feel free to merge as-is.

.github/workflows/update-nock-files.yml
```diff
@@ -49,7 +49,15 @@ jobs:
...
+          sqlite3 tests/nocks.db .dump > tests/nocks.sql
+          echo "result=$(git --no-pager diff --quiet -- tests/nocks.sql || echo "yes")" >> "$GITHUB_OUTPUT"
+          git rm -f tests/nocks.sql
+          mv tests/nocks.db.new tests/nocks.db
```
merceyz, line 60: Can be simplified a bit
```suggestion
        run: |
          sqlite3 tests/nocks.db .dump > /tmp/before.sql
          cp tests/nocks.db tests/nocks.db.new
          git checkout HEAD -- tests/nocks.db
          sqlite3 tests/nocks.db .dump > /tmp/after.sql
          echo "result=$(git --no-pager diff --quiet --no-index /tmp/before.sql /tmp/after.sql || echo "yes")" >> "$GITHUB_OUTPUT"
          mv tests/nocks.db.new tests/nocks.db
```

aduh95:
I confirmed it worked with https://github.com/nodejs/corepack/actions/runs/9400530569/job/25890336262

PR #499, opened 2024-06-11
docs: replace integrity signature algorithm with SHA-512 in README

isudzumi:
After https://github.com/nodejs/corepack/pull/432, looks the hash algorithm for integrity check have switched to SHA-512.
I want to reflect it to README.

### DIFF
--- a/README.md
+++ b/README.md
@@ -73,12 +73,12 @@ Set your package's manager with the `packageManager` field in `package.json`:
 
 ```json
 {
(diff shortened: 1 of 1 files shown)


aduh95:
Not sure I agree with this change, SHA-224 is still a valid algorithm, and before https://github.com/nodejs/corepack/pull/432 Corepack were using SHA-256 anyway. Corepack now defaults to SHA-512 because that's what npm signs, and since we have to calculate the SHA-512 to verify the signature, it's also what we put in the `package.json` – but if the user is providing the SHA, SHA-224 is still a perfectly valid choice.

PR #500, opened 2024-06-12
perf: Improve the performance by running ion concurrent 🚀

sanjaiyan-dev:
This PR uses `Promise.all` to speed things up! 

Currently, some parts of the code wait for each other to finish before starting. `Promise.all` lets them run at the same time, making the app faster.

**Benefits:**

* Blazing fast app (especially for many independent tasks)
* Cleaner code


aduh95:
Thanks for the contribution. This has been superseded by #524

PR #506, opened 2024-06-28
fix: custom registry with path after base url

jorjao81:
when using a custom registry with a path after the base URL (like `https://registry.example.org/with-path/npm`) and using COREPACK_NPM_TOKEN corepack would not set the Bearer token correctly for the file download (it did set it for downloading jsons).

while fixing it I also saw that the logic for setting the bearer token was duplicated. I removed the duplicated logic, always setting the bearer in the same place and removed the strange condition for checking the origin (it is not checked when using username password, why then check when using the token?)

some additional minor changes make the code intent clearer IMHO

let me know if I missed something


aduh95:
Any chance you could instead add a test that uses our custom registry? There are a few unrelated changes in the tests that makes it harder to understand what behavior is being changed

PR #512, opened 2024-07-11
chore: remove direct dependency on tar

wojtekmaj:
`tar` package is not used anywhere in first-party Corepack code. Note that this PR does not remove `tar` from our bundle.

### DIFF
--- a/package.json
+++ b/package.json
@@ -45,7 +45,6 @@
     "proxy-from-env": "^1.1.0",
     "semver": "^7.5.2",
     "supports-color": "^9.0.0",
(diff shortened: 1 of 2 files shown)


merceyz (changes requested):
It's used in multiple places, one of them: https://github.com/nodejs/corepack/blob/09528a8ea8f2a953b67c9079615eae3394531862/sources/corepackUtils.ts#L151-L159.

wojtekmaj:
Oh, I see! The use of backticks fooled me when I was looking for imports. Thanks!

PR #520, opened 2024-07-11
chore: publish to npm with provenance

wojtekmaj:
I find it a bit ironic that corepack checks integrity of downloaded package managers, but corepack itself is published without npm provenance statements. This PR fixes that, leveraging the fact npm is perfectly okay with publishing packages packed using another PM. This is what I did to add npm provenance statements support to dozens of my OSS packages.


.github/workflows/publish.yml
```diff
@@ -44,14 +44,15 @@ jobs:
...
+
+      - run: corepack yarn pack --out corepack.tgz
+
       - name: Publish to the npm registry
-        run: |
-          corepack yarn install --immutable
-          corepack yarn npm publish
+        run: npm publish corepack.tgz --tag latest --provenance
         env:
           YARN_NPM_AUTH_TOKEN: ${{secrets.NPM_TOKEN}}
```
aduh95, line 54:
```suggestion
          NODE_AUTH_TOKEN: ${{secrets.NPM_TOKEN}}
```

.github/workflows/publish.yml
```diff
@@ -44,14 +44,15 @@ jobs:
...
 
+      - run: corepack yarn install --immutable
+
+      - run: corepack yarn pack --out corepack.tgz
+
       - name: Publish to the npm registry
-        run: |
-          corepack yarn install --immutable
-          corepack yarn npm publish
+        run: npm publish corepack.tgz --tag latest --provenance
```
aduh95, line 52: What's the reason to specify `latest` here?

wojtekmaj: You're right - it's latest by default, so it's useless. Removed.

PR #522, opened 2024-07-11
chore: remove supports-color dependency

wojtekmaj:
It's not used directly anywhere.


merceyz:
It's a peer dependency of `debug`.

merceyz (changes requested):
It's a peer dependency of `debug`.

PR #526, opened 2024-07-12
chore: bump Yarn to 4.3.1

aduh95:


yarn.lock
wojtekmaj, line 1: While dependency bumps are welcome, I think they are out of scope described in the PR title - or the other way around, PR title needs to be expanded.

Issue #495, opened 2024-06-10
downloads package managers only from npm


mcollina:
corepack supports downloading the package managers from a source that is not npm. I think this is a significant security risk for our users and we should only download them from npm (or another 3rd party registry, not just a URL).

styfle:
For more context, I created the original issue https://github.com/nodejs/corepack/issues/354 after the Node.js TSC meeting because some expressed that they didn't like the idea of corepack gatekeeping.

mcollina:
The npm client does not download & execute libraries from the internet without telling its users. The gates here should be way tighter to guarantee the basic security of our users.

aduh95:
> The npm client does not download & execute libraries from the internet without telling its users. The gates here should be way tighter to guarantee the basic security of our users.

? That does not seem to be the case

```console
$ mkdir repro && (cd repro && echo '{"dependencies":{"yarn":"https://registry.yarnpkg.com/yarn/-/yarn-1.22.22.tgz"}}' > package.json && npm i) && rm -r repro

added 1 package, and audited 2 packages in 2s

found 0 vulnerabilities
```

No warnings at all, unless it's disabled in my env for some reason. What warning are you seeing?

mcollina:
There are no warnings, but that is ok because the end user asked for it.

Running `yarn` after a `corepack enable` is absolutely _not transparent_ to the end user.
The npm registry acts as a 3rd party that give an extensive list of guarantees. Those guarantees are not available if a project is self-hosting their deps. The user would expect to have the same guarantees as npm.

aduh95:
Oh sorry for the confusion, I thought you were referring to the feature added in https://github.com/nodejs/corepack/pull/359. 
In that case, that's up to the Yarn team (cc @arcanis).
From Corepack side, we could define requirements for what would be acceptable for a default registry – ideally we would not restrict it to npm public registry, but instead define a set of requirements (IIRC we discussed accepting only registries with an audited a signature process)

mcollina:
I agree

merceyz:
> Running yarn after a corepack enable is absolutely not transparent to the end user.

What do you mean?

```console
$ docker run --rm -it node:22 bash
$ corepack enable
$ yarn
! Corepack is about to download https://registry.yarnpkg.com/yarn/-/yarn-1.22.22.tgz
? Do you want to continue? [Y/n]
```

mcollina:
Would the user know the risk this poses? That's my point. The message seems official and safe enough.

[...]

PR #548, opened 2024-08-19
feat!: refuse to download unverified downloads

aduh95:
For the most common use case where a Corepack user interacts with an already-setup-for-Corepack project, this PR doesn't change anything, the project would have a `"packageManager"` field that includes a hash to validate the download.

However, for cases when there are no hash defined there or in the CLI command (`corepack yarn@3.x …`/`corepack up`/`corepack use yarn@3.x`), we need another way to validate the download. For downloads from the npm registry, we verify using the ECDSA signature provided in the package metadata. For package manager that are not downloaded from the npm registry (i.e. Yarn Berry), the download is only protected by TLS, which is pretty good, but not as good.

This change is going to be disruptive to Yarn Berry users. I think the Yarn team can mitigate by either:

- switch to download newer versions of Yarn from the npm registry (we cannot change earlier versions without invalidating everyone's hash).
- send a change for Corepack to validate the download using the npm integrity when no hash is provided.
- provide signatures for Yarn releases in the Yarn registry.

Fixes: https://github.com/nodejs/corepack/issues/495

### DIFF
--- a/sources/corepackUtils.ts
+++ b/sources/corepackUtils.ts
@@ -292,6 +292,8 @@ export async function installVersion(installTarget: string, locator: Locator, {s
       npmRegistryUtils.verifySignature({signatures, integrity, packageName: registry.package, version});
       // @ts-expect-error ignore readonly
       build[1] = Buffer.from(integrity.slice(`sha512-`.length), `base64`).toString(`hex`);
+    } else if (!shouldSkipIntegrityCheck()) {
+      throw new Error(`Cannot find a signature to validate the download. Please provide a hash.`)
     }
   }
   if (build[1] && actualHash !== build[1])
--- a/tests/Up.test.ts
+++ b/tests/Up.test.ts
@@ -14,7 +14,7 @@ describe(`UpCommand`, () => {
   it(`should upgrade the package manager from the current project`, async () => {
     await xfs.mktempPromise(async cwd => {
       await xfs.writeJsonPromise(ppath.join(cwd, `package.json`), {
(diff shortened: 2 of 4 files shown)


sources/corepackUtils.ts
```diff
@@ -292,6 +292,8 @@ export async function installVersion(installTarget: string, locator: Locator, {s
       npmRegistryUtils.verifySignature({signatures, integrity, packageName: registry.package, version});
       // @ts-expect-error ignore readonly
       build[1] = Buffer.from(integrity.slice(`sha512-`.length), `base64`).toString(`hex`);
+    } else if (!shouldSkipIntegrityCheck()) {
+      throw new Error(`Cannot find a signature to validate the download. Please provide a hash.`)
```
aduh95, line 296:
```suggestion
      throw new Error(`Cannot find a signature to validate the download. Please provide a hash.`);
```

arcanis:
> switch to download newer versions of Yarn from the npm registry (we cannot change earlier versions without invalidating everyone's hash).

We're unlikely to do that, as we want our official binaries to be downloaded from our official website.

arcanis (changes requested):
Marking request changes just to be sure this doesn't land before we implement signing in Yarn (which is in our plans, but I'm in paternity leave at the moment).

mcollina (approved):
lgtm

mcollina:
@arcanis I'm glad to hear that signatures are planned for yarn. what's your timeline to add said support? September/October?

tiaraton (approved):

Issue #485, opened 2024-05-28
`COREPACK_ENABLE_AUTO_PIN` should default to `0`


benmccann:
I don't want to use corepack. However, this setting means that every single contributor to my project will need to set `COREPACK_ENABLE_AUTO_PIN=0` and be careful not to accidentally check in the `packageManager` field since corepack will automatically add that field to force itself upon any project you use. This is even more annoying since pnpm can't be used without corepack once the `packageManager` field is present: https://github.com/pnpm/pnpm/issues/8087

Lectem:
Same thing here.
I have projects that use pnpm, some don't, some require pinning the version and others don't.

I think nobody wants their package.json to be automatically modified on any command one could do. And even less on packages you do not own. 

I do NOT want to add this field on packages I do not own just because I contributed to it for a small fix or something.

Discoverability of such a feature is also bad, especially for juniors that will 100% fail to understand the implications of the new field and how to disable the feature.

Lectem:
Note that it can (and should) in my opinion be fixed before the next version of node is released or the problem will most likely blow up.

For reference this was added by #413 in version 0.26. Original issue was #399. 

I think that a warning can be printed by default instead of directly modifying the `package.json`.

AprilNEA:
I also believe that this should be disabled by default.

If we have to synchronize this environment variable within the team to prevent everyone from enabling it, that would be terrible.

KaelWD:
At least allow it to be set in npmrc instead of an environment variable so public projects can force it off.

vtnorton:
For me its another use case. I create a lot or demos, samples and help students with their code. I don't know the environment of each person, and I deal with a lot of different repositories and companies. 

Creating this `packageManager` by default, once I enabled corepack only on my machine, requires everyone to have the exact same package manager (including version) that I have on my computer. This leads to a awful developer experience for trying to running a demo, sample or a open source project for the first time. 

Yes, having an option to disable on `.npmrc` or something would be nice, but in most cases this leads to having more files then I honest need on my repo, that in my use cases I prefer to KISS. 

I did set `COREPACK_ENABLE_AUTO_PIN` to `0` so it doesn't botter me anymore, but I don't think this is the ideal solution.

hardfist:
there're other monorepo tools like Rush which support pin `packageManager` version without using corepack, when run pnpm commands in these tools, the auto pin behavior may make inconsistent version than the version specified by monorepo tools.

hyperknot:
[...]

PR #552, opened 2024-08-23
Do not auto pin by default COREPACK_ENABLE_AUTO_PIN=0 is now the default

Lectem:
This fixes all issues related to auto-pin mentioned in  #485.

### DIFF
--- a/sources/Engine.ts
+++ b/sources/Engine.ts
@@ -263,7 +263,7 @@ export class Engine {
           return fallbackDescriptor;
 
         case `NoSpec`: {
-          if (process.env.COREPACK_ENABLE_AUTO_PIN !== `0`) {
+          if (process.env.COREPACK_ENABLE_AUTO_PIN === `1`) {
             const resolved = await this.resolveDescriptor(fallbackDescriptor, {allowTags: true});
             if (resolved === null)
               throw new UsageError(`Failed to successfully resolve '${fallbackDescriptor.range}' to a valid ${fallbackDescriptor.name} release`);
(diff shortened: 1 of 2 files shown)


sources/Engine.ts
```diff
@@ -275,9 +275,11 @@ export class Engine {
             console.error();
 
             await specUtils.setLocalPackageManager(path.dirname(result.target), installSpec);
+          } else {
+            debugUtils.log(`Falling back to ${fallbackDescriptor.name}@${fallbackDescriptor.range} in the absence of "packageManage" field in ${result.target}.`);
```
styfle, line 279:
```suggestion
            debugUtils.log(`Falling back to ${fallbackDescriptor.name}@${fallbackDescriptor.range} in the absence of "packageManager" field in ${result.target}.`);
```

Lectem:
Github tells me that you requested changes but I'm not sure what changes you asked for ? (maybe the comments were not commited, I've seen this happen in the past if you don't submit the review)

> Without auto-pinning, corepack is the source of significant bugs and failed builds, as the version of the package manager is essentially free floating.

While I do agree, this is very invasive as mentioned in the issue. To sum them up again here:
- Any contributor using corepack will need to either ignore the `package.json` changes or try to submit their own version of the packagemanager field, which may not be well tested by the maintainers.
- Corepack should not attempt to write read-only files, this defeats the whole purpose of some command parameters of the package managers such as `npm ci` `npm install -save=false` and `pnpm --frozen-lockfile`. I should be able to build without triggering "failed to write" issues.

That is why I made it so that a warning will be logged as long as the user does not provide the `packageManager` field. If it needs more discoverability, we could output it to stderr instead of the logs ?

> corepack is the source of significant bugs and failed builds

I mean, that's not any different from people installing their own version of npm, pnpm or yarn. If the package.json does not restrict the package manager version properly, it's an issue whether or not you use corepack.

Lectem:
An alternative to changing `COREPACK_ENABLE_AUTO_PIN` default, would be to actually prompt the user if they actually want you to do the change.
In general, modifying (especially source) files without user consent is a bad idea IMHO.

So what one could do:
- if user has a non redirected console (basicaly `noTTY() && !CI)`) => prompt them to apply the change
- if no input or CI or file readonly => do nothing except if `COREPACK_ENABLE_AUTO_PIN` is set (to either `0` or `1` explicitly)

mcollina:
> I mean, that's not any different from people installing their own version of npm, pnpm or yarn. If the package.json does not restrict the package manager version properly, it's an issue whether or not you use corepack.

Indeed it's the same. The change was requested when it was asked to enable corepack by default with Node.js.

In that specific instance, there is _no user will to install a package manager_.

If the user does not know that they are actually installing a package manager (and what version that is), I think the only sensible thing to do is to add the packageManager field.
 
---

My -1 is that I don't think this should land at all in this form.

PR #574, opened 2024-11-17
perf: prefer `module.enableCompileCache` over `v8-compile-cache`

merceyz:


sources/corepackUtils.ts
```diff
@@ -400,12 +400,16 @@ export async function runVersion(locator: Locator, installSpec: InstallSpec & {s
...
-  // [SIGSEGV]
-  if (locator.name !== `npm` || semverLt(locator.reference, `9.7.0`))
-    // @ts-expect-error - No types
-    await import(`v8-compile-cache`);
+  // @ts-expect-error - Missing types
+  if (!Module.enableCompileCache) {
+    // Node.js segfaults when using npm@>=9.7.0 and v8-compile-cache
+    // $ docker run -it node:20.3.0-slim corepack npm@9.7.1 --version
+    // [SIGSEGV]
+    if (locator.name !== `npm` || semverLt(locator.reference, `9.7.0`)) {
```
aduh95, line 408: nit: we can minimize the diff, to avoid cluttering git blame
```suggestion
  // Node.js segfaults when using npm@>=9.7.0 and v8-compile-cache
  // $ docker run -it node:20.3.0-slim corepack npm@9.7.1 --version
  // [SIGSEGV]
  if (locator.name !== `npm` || semverLt(locator.reference, `9.7.0`)) {
    // @ts-expect-error - Missing types
    if (!Module.enableCompileCache) {
```

merceyz: The suggested change would exclude npm from the code cache which we don't need to do.

merceyz: I haven't benchmarked it but this would cause it to do work it doesn't have to (semverLt)

aduh95: Only for npm though

merceyz: It would always hit npm as it has to check the version.

aduh95: Yeah that's what I'm saying, and if native support is indeed disabled – otherwise it would have to do the check anyway

Issue #592, opened 2025-01-13
Update better-sqlite3 to >= 11.5.0


MikeMcC399:
- Supersedes https://github.com/nodejs/corepack/issues/591

## What would you like?

1. Update [better-sqlite3](https://www.npmjs.com/package/better-sqlite3) to >= [better-sqlite3@11.5.0](https://github.com/WiseLibs/better-sqlite3/releases/tag/v11.5.0). Latest version is currently [better-sqlite3@11.7.2](https://github.com/WiseLibs/better-sqlite3/releases/tag/v11.7.2)
2. Drop support for Node.js 21 ([package.json](https://github.com/nodejs/corepack/blob/main/package.json) `engines` and [ci.yml](https://github.com/nodejs/corepack/blob/main/.github/workflows/ci.yml))
3. Add CI testing in Node.js 23 ([ci.yml](https://github.com/nodejs/corepack/blob/main/.github/workflows/ci.yml))

## Why is this needed?

1. Node.js 23 was [released](https://github.com/nodejs/release#release-schedule) on Oct 15, 2024 and is the "current" version. It should be tested in CI. [Node.js v23.6.0 documentation](https://nodejs.org/docs/latest/api/corepack.html) includes it in the `corepack` section of the documentation.
2. [better-sqlite3](https://www.npmjs.com/package/better-sqlite3) support for Node.js 23 was introduced in [better-sqlite3@11.5.0](https://github.com/WiseLibs/better-sqlite3/releases/tag/v11.5.0). This is the minimum version required to run `yarn install` successfully with Node.js 23. `yarn install` is a prerequisite to run `yarn build` and `yarn test`.
3. [better-sqlite3@11.0.0](https://github.com/WiseLibs/better-sqlite3/releases/tag/v11.0.0) dropped support for Node.js 21
4. There is no version of [better-sqlite3](https://www.npmjs.com/package/better-sqlite3) which supports both Node.js 21 and 23 (see also comments in https://github.com/nodejs/corepack/pull/513 from @wojtekmaj).
5. Node.js 21 entered [end-of-life](https://github.com/nodejs/release#release-schedule) on Jun 6, 2024. It should no longer be tested in CI since it is now unsupported.

## Other

PRs

1. [x] Drop Node.js 21 support
2. [x] Update better-sqlite3 to >= 11.5.0

MikeMcC399:
- See #594

PR #594, opened 2025-01-14
chore!: remove Node.js 21.x from the range of supported versions

MikeMcC399:
- closes https://github.com/nodejs/corepack/issues/592

## Issue

1. Node.js 21 entered [end-of-life](https://github.com/nodejs/release#release-schedule) on Jun 6, 2024. It is however still allowed by:

https://github.com/nodejs/corepack/blob/0ae3c3c1f5cbbc7fd14b7be8e71e8472215c68d2/package.json#L12-L14

2. The workflow [ci.yml](https://github.com/nodejs/corepack/blob/main/.github/workflows/ci.yml) continues to test it:

https://github.com/nodejs/corepack/blob/0ae3c3c1f5cbbc7fd14b7be8e71e8472215c68d2/.github/workflows/ci.yml#L47-L52

3. There is no version of [better-sqlite3](https://www.npmjs.com/package/better-sqlite3) (used in `devDependencies`) which supports both Node.js 21 and 23, so continuing support for Node.js 21 blocks supporting Node.js 23.

## Change

- Drop support for Node.js 21 in [package.json](https://github.com/nodejs/corepack/blob/main/package.json) `engines` and testing in [ci.yml](https://github.com/nodejs/corepack/blob/main/.github/workflows/ci.yml).
- Update [better-sqlite3](https://www.npmjs.com/package/better-sqlite3) to [better-sqlite3@11.7.2](https://github.com/WiseLibs/better-sqlite3/releases/tag/v11.7.2) (Edit: was `latest` when this PR was originally submitted).

### Additional change
- Add Node.js 23 to workflow and exclude `windows-latest` from testing due to issue #597

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -49,10 +49,18 @@ jobs:
           - 18
           - 20
           - 22
+          - 23
         platform:
           - ubuntu-latest
           - macos-latest
           - windows-latest
+
+      # Temporarily skipping Node.js 23 under Windows due to issue
+      # https://github.com/nodejs/corepack/issues/597
+      # ci vitest fails "handle integrity checks" on Windows Node.js 23
+        exclude:
+          - node: 23
+            platform: windows-latest
 
     name: "${{matrix.platform}} w/ Node.js ${{matrix.node}}.x"
     runs-on: ${{matrix.platform}}
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -48,7 +48,6 @@ jobs:
         node:
           - 18
           - 20
-          - 21
           - 22
         platform:
           - ubuntu-latest
--- a/package.json
+++ b/package.json
@@ -26,7 +26,7 @@
     "@yarnpkg/eslint-config": "^2.0.0",
     "@yarnpkg/fslib": "^3.0.0-rc.48",
     "@zkochan/cmd-shim": "^6.0.0",
-    "better-sqlite3": "^10.0.0",
+    "better-sqlite3": "^11.7.2",
     "clipanion": "patch:clipanion@npm%3A3.2.1#~/.yarn/patches/clipanion-npm-3.2.1-fc9187f56c.patch",
     "debug": "^4.1.1",
     "esbuild": "^0.21.0",
--- a/package.json
+++ b/package.json
@@ -10,7 +10,7 @@
     "url": "https://github.com/nodejs/corepack.git"
   },
   "engines": {
-    "node": "^18.17.1 || >=20.10.0"
+    "node": "^18.17.1 || ^20.10.0 || >=22.11.0"
   },
   "exports": {
     "./package.json": "./package.json"
--- a/yarn.lock
+++ b/yarn.lock
@@ -991,14 +991,14 @@ __metadata:
   languageName: node
   linkType: hard
 
-"better-sqlite3@npm:^10.0.0":
-  version: 10.1.0
-  resolution: "better-sqlite3@npm:10.1.0"
+"better-sqlite3@npm:^11.7.2":
+  version: 11.7.2
+  resolution: "better-sqlite3@npm:11.7.2"
   dependencies:
     bindings: "npm:^1.5.0"
     node-gyp: "npm:latest"
     prebuild-install: "npm:^7.1.1"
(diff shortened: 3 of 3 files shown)


.github/workflows/ci.yml
```diff
@@ -48,7 +48,6 @@ jobs:
         node:
           - 18
           - 20
-          - 21
           - 22
```
aduh95, line 51: It looks like a good opportunity to add 23.x to the matrix, wdyt?
```suggestion
          - 22
          - 23
```

aduh95: Why not update better-sqlite in the same PR?

MikeMcC399: The alternative would be one PR to
- drop Node 21
- upgrade better-sqlite2
- add Node 23
all together

MikeMcC399: OK, I will add the additional commits to this PR, if that is what you prefer.

MikeMcC399: Done

package.json
```diff
@@ -10,7 +10,7 @@
     "url": "https://github.com/nodejs/corepack.git"
   },
   "engines": {
-    "node": "^18.17.1 || >=20.10.0"
+    "node": "^18.17.1 || ^20.10.0 || >=22.0.0"
```
aduh95, line 13: nit: we could restrict it to the first LTS version of 22.x
```suggestion
    "node": "^18.17.1 || ^20.10.0 || >=22.10.0"
```

aduh95 (approved):

MikeMcC399:
@aduh95 

> It looks like the exit code is different on Windows with Node.js 23.x for some reason 🤔

That is going to need separate investigation. The test appears to be flaky, as I had previously seen it successful.

I am going to remove Node.js 23 tests from this PR so it should then pass all tests. Stand by!

aduh95:
I don't think removing 23.x is the correct way forward, it looks like it would be more reasonable to either skip this particular test on 23.x, or account for the other exit code. AFAICT it's not flaky on CI

MikeMcC399:
@aduh95 

> I don't think removing 23.x is the correct way forward, it looks like it would be more reasonable to either skip this particular test on 23.x, or account for the other exit code. AFAICT it's not flaky on CI, it looks like it's consistently failing with the same error

I talking about moving the update to test in 23.x into a separate PR and this is a different issue which needs to be resolved.

MikeMcC399:
@aduh95 

I suggest to merge this now, since it cleanly passes CI. It's a good basis now for working out what to do with Node.js 23 in a separate PR.

MikeMcC399:
@aduh95 

- I've logged the Windows / Node.js 23.x test failure issue in https://github.com/nodejs/corepack/issues/597 since I was able to reproduce it locally. I don't have any explanation why it worked once in CI and is now failing all the time. If it's now reproducible, it should make it easier to deal with it anyway.

aduh95:
> I talking about moving the update to test in 23.x into a separate PR and this is a different issue which needs to be resolved.

Is it a different issue? It seems to me that it's very much related, we're dropping 21.x support because we want to support 23.x and there's no other solution because of what supports better-sqlite3.
I think I would prefer we address it in this PR, either by editing the test and adding a TODO, or by fixing the underlying issue if we're able to figure it out

MikeMcC399:
@aduh95 

- You are right that the motivation was to test in Node.js 23, however dropping Node.js 21 should be done in any case since it is EOL. That was done separately for Node.js 19 in PR #375.

- I would prefer to merge this PR as is. I generally prefer smaller self-contained PRs.

- Your preference is different and involves solving #597 as a prerequisite. It's your decision of course. At this point in time I don't think I can contribute much more however.

MikeMcC399:
@aduh95 

- Thank you for progressing the issue https://github.com/nodejs/corepack/issues/597! At this time I would assume that the test problem in corepack is caused by a regression in Node.js 23.0.0 for Windows, that it will be fixed in Node.js and that it will need at least a few weeks to get a fix into a new Node.js 23.x release.

In the meantime I propose to add Node.js 23 to the test matrix for corepack and temporarily exclude testing on Windows with Node.js 23. I think this is the cleanest solution.  Windows is already being tested on other Node.js versions (18, 20 & 22), so the testing gap is minimal.

The advantage is that introducing tests with Node.js 23.x for Ubuntu and macOS is not delayed further.

Once the Windows issue is solved, it is easy to remove the exclusion.

I believe this PR is now in a state where you could consider merging it.

.github/workflows/ci.yml
MikeMcC399: Node.js 23 is now added again

aduh95 (approved):

Issue #598, opened 2025-01-15
Contributor Windows build on non-HOMEDRIVE 'Could not resolve "clipanion"'


MikeMcC399:
## Issue

With a clone of https://github.com/nodejs/corepack on a Windows 11 `D:` drive `corepack yarn build` fails with:

> ✘ [ERROR] Could not resolve "clipanion"

and other similar errors.

I ran into this issue when following up testing on Windows. If it is an expected restriction, then it could be added as a note to the [CONTRIBUTING](https://github.com/nodejs/corepack/blob/main/CONTRIBUTING.md) document.

## Steps to reproduce

On Windows 11 24H2 with Node.js `22.13.0` LTS:

```shell
git clone https://github.com/nodejs/corepack # to drive which is not the Windows `HOMEDRIVE`
cd corepack
corepack yarn install
corepack yarn build
```

## Logs

```text
$ corepack yarn install
corepack yarn build
➤ YN0000: · Yarn 4.6.0
➤ YN0000: ┌ Resolution step
➤ YN0000: └ Completed
➤ YN0000: ┌ Post-resolution validation
➤ YN0086: │ Some peer dependencies are incorrectly met by dependencies; run yarn explain peer-requirements for details.
➤ YN0000: └ Completed
➤ YN0000: ┌ Fetch step
➤ YN0013: │ 424 packages were added to the project (+ 131.97 MiB).
➤ YN0000: └ Completed in 10s 889ms
➤ YN0000: ┌ Link step
➤ YN0000: │ ESM support for PnP uses the experimental loader API and is therefore experimental
➤ YN0000: └ Completed
➤ YN0000: · Done with warnings in 11s 136ms
✘ [ERROR] Could not resolve "clipanion"

    sources/main.ts:1:44:
      1 │ import {BaseContext, Builtins, Cli}    from 'clipanion';
        ╵                                             ~~~~~~~~~~~

  You can mark the path "clipanion" as external to exclude it from the bundle, which will remove
  this error and leave the unresolved path in the bundle.

✘ [ERROR] Could not resolve "clipanion"

    sources/commands/deprecated/Prepare.ts:1:42:
      1 │ import {Command, Option, UsageError} from 'clipanion';
        ╵                                           ~~~~~~~~~~~

  You can mark the path "clipanion" as external to exclude it from the bundle, which will remove
  this error and leave the unresolved path in the bundle.

✘ [ERROR] Could not resolve "@zkochan/cmd-shim"

[...]

PR #606, opened 2025-01-21
docs: add contributor windows os non-homedrive usage

MikeMcC399:
- addresses documentation related to #598

## Issue

- As described in issue https://github.com/nodejs/corepack/issues/598, if the instructions in [CONTRIBUTING](https://github.com/nodejs/corepack/blob/main/CONTRIBUTING.md) are followed to clone the repo, install dependencies with Yarn and then build with Yarn, these steps fail if running on a Microsoft Windows operating system and the repo has been cloned to a drive which is not the user's `HOMEDRIVE`.

- Pending the implementation of Yarn feature request https://github.com/yarnpkg/berry/issues/4030, [Yarn Plug'n'Play](https://yarnpkg.com/features/pnp) is not compatible with the above Windows configuration.

- The [corepack repo](https://github.com/nodejs/corepack) is currently configured to use [Yarn 4.3.1](https://github.com/yarnpkg/berry/releases/tag/%40yarnpkg%2Fcli%2F4.3.1) and since [nodeLinker](https://yarnpkg.com/configuration/yarnrc#nodeLinker) is unconfigured, it defaults to `nodeLinker: "pnp"` for [Yarn Plug'n'Play](https://yarnpkg.com/features/pnp).

https://github.com/nodejs/corepack/blob/2a04e7c8da26807d36825cb289d5d9d3a0c71669/package.json#L19

## Change

Add a recommendation and workaround instructions to [CONTRIBUTING](https://github.com/nodejs/corepack/blob/main/CONTRIBUTING.md) for contributors working on Microsoft Windows, related to drive usage.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -2,7 +2,7 @@
 
 If you want to build Corepack yourself, you can build the project like this:
 
-1. Clone this repository. (See below [Contributing on Windows](#contributing-on-windows) if you plan to use Microsoft Windows.)
+1. Clone this repository. Some restrictions on the cloning location apply to Windows users, read [Contributing on Windows](#contributing-on-windows) for more information.
 2. Run `yarn install` (or `corepack yarn install` if the global version of
    `yarn` is not provided by Corepack).
 3. Run `yarn build` (or `corepack yarn build`).
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -2,7 +2,7 @@
 
 If you want to build Corepack yourself, you can build the project like this:
 
-1. Clone this repository.
+1. Clone this repository. (See below [Contributing on Windows](#contributing-on-windows) if you plan to use Microsoft Windows.)
 2. Run `yarn install` (or `corepack yarn install` if the global version of
    `yarn` is not provided by Corepack).
 3. Run `yarn build` (or `corepack yarn build`).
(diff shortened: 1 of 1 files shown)


MikeMcC399:
Preview

https://github.com/MikeMcC399/corepack/blob/docs/windows-drive/CONTRIBUTING.md

CONTRIBUTING.md
```diff
@@ -2,7 +2,7 @@
 
 If you want to build Corepack yourself, you can build the project like this:
 
-1. Clone this repository.
+1. Clone this repository. (See below [Contributing on Windows](#contributing-on-windows) if you plan to use Microsoft Windows.)
```
aduh95, line 5:
```suggestion
1. Clone this repository. N.B.: Windows users might need to read the [Contributing on Windows](#contributing-on-windows) section before cloning.
```

MikeMcC399: I prefer the original text since Windows users will need to read the text first to find out if it affects them, and I wouldn't suggest not reading it, so saying "might need to read" implies that you could ignore the section first.

I ran into the issue on my desktop test machine which has an SSD C: drive and a separate hard drive D: as standard. I cloned into the D: drive where I have more space and that's when I hit this issue.

The issue could be avoided also if the Corepack repo used `nodeLinker: node-modules` instead of the default `pnp`, in which case this special section for Windows users would not be necessary. That is something which could be considered separately.

Also, as a secondary point, the [Microsoft Style Guide](https://learn.microsoft.com/en-us/style-guide/word-choice/use-us-spelling-avoid-non-english-words) advises "Avoid Latin abbreviations for common English phrases." so I would not use "N.B.".

If you feel strongly that the text should be changed to your suggestion, then I would merge it, however I would like to keep what I originally wrote.

aduh95: There are two things I don't like in the current wording:

- "if you plan to use", if you're working on Linux and "plan to use" Windows, I don't see how it's relevant.
- "Microsoft Windows", that's just a weird way of talking about it
```suggestion
1. Clone this repository. Some restrictions on the cloning location apply to Windows users, read [Contributing on Windows](#contributing-on-windows) for more information.
```

Issue #618, opened 2025-01-31
Outdated pnpm 4.x - 6.x versions tested


MikeMcC399:
## Issue

Versions of pnpm `5.x` - `6.x` used in [tests/main.test.ts](https://github.com/nodejs/corepack/blob/main/tests/main.test.ts), and supposed to be active, are no longer supported by pnpm. Conversely, the versions of pnpm which are currently supported are not tested.

### Corepack

Corepack [tests](https://github.com/nodejs/corepack/tree/main/tests) use:

- `pnpm@5.8.0`
- `pnpm@4.11.6`
- `pnpm@6.6.2`

### pnpm

The [pnpm documentation website](https://pnpm.io/) displays documentation for versions `8.x`, `9.x` and `10.x` only. According to [pnpm Security Policy](https://github.com/pnpm/pnpm/blob/main/SECURITY.md) version `7.x` and lower are unsupported.

## Suggestion

Update [tests/main.test.ts](https://github.com/nodejs/corepack/blob/main/tests/main.test.ts) to remove unsupported pnpm versions and add supported versions.

Use the values from:

```shell
npm view pnpm dist-tags
```

which are currently:

```text
  'latest-8': '8.15.9',
  'latest-9': '9.15.4',
  'latest-10': '10.1.0',
```

aduh95:
I don't think it's a problem that Corepack supports loading EOL versions of pnpm, we're not going to tell users which version to use.

MikeMcC399:
@aduh95 

Based on the use of one npm keyid to sign all pnpm releases up until the last few days, I am OK with closing this issue.

> I don't think it's a problem that Corepack supports loading EOL versions of pnpm, we're not going to tell users which version to use.

Possibly there should be some definition of a separate support policy for Corepack including some language about whether package managers which have been declared EOL and unsupported by their package manager maintainer authority are still supported for installation by Corepack or not. I was assuming that they would not be supported by Corepack, however this should be a separate discussion, not in this issue.

PR #621, opened 2025-02-01
chore(test): update pnpm tests to use current version of the package manger

joemckenney:
Follow suggestions for fix in https://github.com/nodejs/corepack/issues/618

### DIFF
--- a/tests/main.test.ts
+++ b/tests/main.test.ts
@@ -120,12 +120,9 @@ const testedPackageManagers: Array<[string, string] | [string, string, string]> 
   [`yarn`, `3.0.0-rc.2`],
   [`yarn`, `3.0.0-rc.2+sha1.694bdad81703169e203febd57f9dc97d3be867bd`],
   [`yarn`, `3.0.0-rc.2+sha224.f83f6d1cbfac10ba6b516a62ccd2a72ccd857aa6c514d1cd7185ec60`],
-  [`pnpm`, `4.11.6`],
-  [`pnpm`, `4.11.6+sha1.7cffc04295f4db4740225c6c37cc345eb923c06a`],
-  [`pnpm`, `4.11.6+sha224.7783c4b01916b7a69e6ff05d328df6f83cb7f127e9c96be88739386d`],
-  [`pnpm`, `6.6.2`],
-  [`pnpm`, `6.6.2+sha1.7b4d6b176c1b93b5670ed94c24babb7d80c13854`],
-  [`pnpm`, `6.6.2+sha224.eb5c0acad3b0f40ecdaa2db9aa5a73134ad256e17e22d1419a2ab073`],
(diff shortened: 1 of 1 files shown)


MikeMcC399:
@joemckenney 

- I was actually waiting for some feedback from the maintainers on issue #618 before taking any steps to submit a PR myself. I guess I should have written that in the issue!
- In the meantime I played around with the tests a little, and found that `nock.db`  needs to be deleted and recreated. This PR fails in GitHub Actions for that reason.
- This PR isn't testing pnpm with hashes. That is probably something that the maintainers would want to test, however I can't speak for them. Probably SHA1 should be tested. I don't know about continuing to test SHA224 though, since latest Corepack versions no longer create SHA224, although they still accept SHA224. Probably SHA512 should be tested in addition.

joemckenney:
> * In the meantime I played around with the tests a little, and found that `nock.db`  needs to be deleted and recreated. This PR fails in GitHub Actions for that reason.
> * This PR isn't testing pnpm with hashes. That is probably something that the maintainers would want to test, however I can't speak for them. Probably SHA1 should be tested. I don't know about continuing to test SHA224 though, since latest Corepack versions no longer create SHA224, although they still accept SHA224. Probably SHA512 should be tested in addition.

Happy to take on both of these.  Note that I don't see the failing GHA workflows on my end, presumably because I don't have permissions and/or i'm not a member?

MikeMcC399:
@joemckenney 

- Thanks for your willingness to work together!

- The suggestion to update these versions was however rejected in https://github.com/nodejs/corepack/issues/618. Without sponsorship from a maintainer I would not expect this PR to be approved. So unless there is a change in direction from the maintainers I suggest not to invest more effort into this PR.

- You don't see your workflow failing because your PR checks haven't been approved to run. I copied your branch `joemckenney:joe/corepack/issues/618` to my fork and ran it there. You can see the failing log under https://github.com/MikeMcC399/corepack/actions/runs/13086183011

aduh95:
A PR that increases the coverage the coverage (i.e. add new versions to test with) would probably be accepted. However, this one decreases the coverage (IMO it's actually good that we test Corepack still works for older releases), so I don't think that would be an improvement as is.

I share the sentiment that it is not worth investing your time – those versions are not special, if Corepack breaks with those one, it would also break with the one being tested. Since we test latest versions when they come out already, if seems to me this PR would only increase the maintenance burden (we'd have to send PRs everytime another release branch of pnpm is EOL / semver is cutoff) while improving Corepack reliability only marginally at best. If you were to send a PR that adds a workflow that somehow updates our tests, now we'd be talking, but it's up to you to decide whether it is worth investing the time to build such workflow for ultimately fixing a (IMO) non-issue.

MikeMcC399:
@joemckenney 

I suggest to close this issue.

Issue #567, opened 2024-10-08
Implement devEngines proposal from Package Metadata Interoperability Collab Space


trivikr:
### Is your feature request related to a problem? Please describe.

The Package Metadata Interoperability Collab Space has a consensus on devEngines proposal, and it was merged in https://github.com/openjs-foundation/package-metadata-interoperability-collab-space/pull/27

https://github.com/npm/cli/releases/tag/v10.9.0 has an initial implementation of this spec, as the following is an example user experience:

```console
$ test-devEngines> npm -v
10.9.0

$ test-devEngines> echo '{ "devEngines": { "packageManager": { "name": "yarn" } } }' > package.json

$ test-devEngines> npm install
npm error code EBADDEVENGINES
npm error EBADDEVENGINES The developer of this package has specified the following through devEngines
npm error EBADDEVENGINES Invalid engine "packageManager"
npm error EBADDEVENGINES Invalid name "yarn" does not match "npm" for "packageManager"
npm error EBADDEVENGINES {
npm error EBADDEVENGINES   current: { name: 'npm', version: '10.9.0' },
npm error EBADDEVENGINES   required: { name: 'yarn' }
npm error EBADDEVENGINES }
```

Although npm@10.9.0 implements this, the proposal doesn't explicitly mention whose responsibility it should be at the time of comment.

### Describe the solution you'd like

Corepack implements devEngines proposal, so that existing users can continue using corepack.

### Describe alternatives you've considered
Request my package manager to implement devEngines, and update my dependencies to that version

aduh95:
> Corepack implements devEngines proposal, so that existing users can continue using corepack.

Can you clarify what you mean by that? How/why would existing users not be able to continue using Corepack with devEngines?

aduh95:
> Corepack implements devEngines proposal, so that existing users can continue using corepack.

Can you clarify what you mean by that? How/why would existing users not be able to continue using Corepack with devEngines?

trivikr:
Since `devEngines` is proposed by Package Metadata Interoperability Collab Space and implemented by npm, it appears to be a standard over existing `packageManager` field for an average developer in JS ecosystem. If that's the case, it means the ecosystem should move towards `devEngines` over long term.

I see it as CJS and ESM. Node.js adopted CJS module system in 2009 as there was no standard in JS ecosystem. But after ESM was standardized in ES2015, Node.js gradually added support for it and now the ecosystem is slowly moving towards ESM.

[...]

Issue #402, opened 2024-02-27
`packageManager` field is too limited


GeoffreyBooth:
The `packageManager` field as currently designed is too limited for what I want to do:

- As a project author, I want to enforce a version range, like npm >= 10.0.0.

- As a project author, I want to control what happens when the developer uses a package manager that fails validation, either because it’s the wrong package manager or a version that’s outside of the defined range:
  - Should it warn, but then proceed
  - Should it warn, and then prompt to download the newest supported version, and then try again in that version
  - Should it error, and exit (like `engines.strict`)

- As a project author, I want to control over package managers based on operations. For example, I want to ensure that `install` commands use the package manager I specify, but `run` commands can use any package manager. (So install via `pnpm` and run scripts via Bun, for example.)

Seeing as this much complexity will require multiple configuration options, I suggest we retire the `packageManager` field and create a new `corepack` field that can contain a configuration object with all the necessary fields to cover the various permutations described above.

@nodejs/corepack @nodejs/loaders @nodejs/npm @nodejs/package-maintenance

arcanis:
This issue is about several independent feature requests ("I want to use different package managers on different commands", "I want to specify what happens when the version mismatches"). The `packageManager` field itself has no impact on those, and it makes little sense to start the discussion from that.

Can you close this thread and open dedicated ones for each of those two use cases?

GeoffreyBooth:
The request is that I want more configuration options for Corepack. Specifically, these are the things I want to be able to configure. I can split it into multiple issues, but I think these all go together (many of these ideas are current features of the `engines` field) and the question is really “what use cases do we want to support” which would lead to an update of the design doc (#401).

aduh95:
`packageManager` is not a replacement for the `engines` field, its goal is more narrow (and that’s probably a good thing)

GeoffreyBooth:
> is not a replacement for the engines field

It should be. The `engines` field can't do some of these things, and for the parts that overlap there should be a clear precedence rather than the two being mixed.

GeoffreyBooth:
> its goal is more narrow 

Where is the goal defined?

aduh95:
> > its goal is more narrow
> 
> Where is the goal defined?

https://nodejs.org/api/packages.html#packagemanager

GeoffreyBooth:
[...]

Issue #628, opened 2025-02-03
COREPACK_ENABLE_AUTO_PIN should be possible to set on a project level


SimonMorkSaetre:
I encountered an issue where the packageManager field was added every single time I installed with yarn. We don't allow the field to be added to our package.json (I have tried but it's a management problem), so the field has to be removed manually every time one reinstalls packages. Yes one can add an environment variable on an OS level, but it has to be done on every single device that is being used with our codebase. It adds a lot of extra work for us.

Please either disable the packageManager field by default or add a way to disable it on a project level.

MikeMcC399:
- related to https://github.com/nodejs/corepack/issues/485

mjracy:
[向日葵远程控制_Main-1739200701386.zip](https://github.com/user-attachments/files/18736844/_Main-1739200701386.zip)

PR #634, opened 2025-02-06
feat: add limited support for `devEngines` and `.corepack.env`

aduh95:
This PR introduces the concept of `.corepack.env` which would allow project maintainers to customize some parts of Corepack. ~We realistically cannot release it until Node.js 18.x is EOL~ (env parser is only available on Node.js 20.x LTS and later) in April, EDIT: actually it won't be as big of a deal as I thought, it falls back elegantly when run on a Node.js version without `.env` support.

`devEngines` support in this PR is limited to `packageManager` only, only one `packageManager` can be specified, the `onFail` field is ignored.
~This PR also does not add support of `devEngines` to `corepack use` and `corepack up`, which should be addressed before releasing.~ EDIT: this has been addressed

Fixes: https://github.com/nodejs/corepack/issues/567
Fixes: #402 
Fixes: #95 

Should unblock fixing: https://github.com/nodejs/corepack/issues/628

### DIFF
--- a/README.md
+++ b/README.md
@@ -113,6 +113,35 @@ use in the archive).
 }
 ```
 
+#### `devEngines.packageManager`
+
+When a `devEngines.packageManager` field is defined, and is an object containing
+a `"name"` field (can also optionally contain `version` and `onFail` fields),
+Corepack will use it to validate you're using a compatible package manager.
+
+Depending on the value of `devEngines.packageManager.onFail`:
+
+- if set to `ignore`, Corepack won't print any warning or error.
+- if unset or set to `error`, Corepack will throw an error in case of a mismatch.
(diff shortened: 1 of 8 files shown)


styfle (commented):
Can we separate into two PRs since this sounds like two separate features?

- add support for `devEngines`
- add support for `.corepack.env`

This way we don't block the devEngines feature on account of any bikesheding that could come from `.corepack.env`

GeoffreyBooth:
> devEngines support in this PR is limited to packageManager only, only one packageManager can be specified, the onFail field is ignored.

Can you please explain these limitations?

aduh95:
Superseded by https://github.com/nodejs/corepack/pull/642, https://github.com/nodejs/corepack/pull/643, and https://github.com/nodejs/corepack/pull/668

Issue #635, opened 2025-02-06
proxy quirks


aep-sunlife:
* I don't see the [node-proxy-agent](https://github.com/TooTallNate/node-proxy-agent) dependency registered in package.json. Is this truly the HTTP client framework that corepack uses to download components when running `corepack install`?
* node-proxy-agent is a dead project with a successor "proxy-agents". When can we replace the dead proxy dependency with a live one?
* Despite both corepack and node-proxy-agent documenting uppercase `HTTP_PROXY`, `HTTPS_PROXY` environment variables, on my machine only the lowercase ones `http_proxy`, `https_proxy` have any effect on whether package installation occurs through a proxy or not.
* Unlike many other applications, including NPM, corepack fails when presented with the proxy URL's `http://` or `https://` (no address). The user's intention here is to indicate that no proxy should be used. And these URL values make larger configuration scripting a lot easier.
* Why not use the same HTTP client as NPM, so that proxy configuration at least behaves identically? Then any questions about configuration can simply link to official NPM documentation. Smaller technology stack, smaller attack surface, less maintenance.

aduh95:
> Why not use the same HTTP client as NPM […] Smaller technology stack, smaller attack surface

I very much doubt using npm stack would decrease the size of bundle and the attack surface, quite the contrary.



> * I don't see the [node-proxy-agent](https://github.com/TooTallNate/node-proxy-agent) dependency registered in package.json. Is this truly the HTTP client framework that corepack uses to download components when running `corepack install`?

`corepack install` is not a thing (yet), but no, we moved away from it with https://github.com/nodejs/corepack/pull/365.
I've opened #636 to address this.

> * Despite both corepack and node-proxy-agent documenting uppercase `HTTP_PROXY`, `HTTPS_PROXY` environment variables, on my machine only the lowercase ones `http_proxy`, `https_proxy` have any effect on whether package installation occurs through a proxy or not.
> * Unlike many other applications, including NPM, corepack fails when presented with the proxy URL's `http://` or `https://` (no address). The user's intention here is to indicate that no proxy should be used. And these URL values make larger configuration scripting a lot easier.

That feels something that should be reported upstream, but I'm not opposed to fix it downstream if someone sends a PR.

aduh95:
> * Unlike many other applications, including NPM, corepack fails when presented with the proxy URL's `http://` or `https://` (no address). The user's intention here is to indicate that no proxy should be used. And these URL values make larger configuration scripting a lot easier.

[...]

PR #636, opened 2025-02-06
fix(doc): fix link to proxy library

aduh95:
Fixes: #635

### DIFF
--- a/README.md
+++ b/README.md
@@ -294,7 +294,7 @@ same major line. Should you need to upgrade to a new major, use an explicit
   empty password, explicitly set `COREPACK_NPM_PASSWORD` to an empty string.
 
 - `HTTP_PROXY`, `HTTPS_PROXY`, and `NO_PROXY` are supported through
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -294,7 +294,7 @@ same major line. Should you need to upgrade to a new major, use an explicit
   empty password, explicitly set `COREPACK_NPM_PASSWORD` to an empty string.
 
 - `HTTP_PROXY`, `HTTPS_PROXY`, and `NO_PROXY` are supported through
-  [`node-proxy-agent`](https://github.com/TooTallNate/node-proxy-agent).
+  [`proxy-from-env`](https://github.com/Rob--W/proxy-from-env).
```
merceyz, line 297: I think we should be a bit more specific here.
We use `proxy-from-env` to get which proxy address to pass to Undici.

aduh95: Does it matter?

merceyz: Not really, no.
Feel free to ignore this.

PR #640, opened 2025-02-07
fix: switch from `proxy-from-env` to `EnvHttpProxyAgent`

### LINKED ISSUE #635: proxy quirks

aduh95:
Fixes: https://github.com/nodejs/corepack/issues/635

### DIFF
--- a/README.md
+++ b/README.md
@@ -293,8 +293,8 @@ same major line. Should you need to upgrade to a new major, use an explicit
   environment variables are required and as plain text. If you want to send an
   empty password, explicitly set `COREPACK_NPM_PASSWORD` to an empty string.
 
-- `HTTP_PROXY`, `HTTPS_PROXY`, and `NO_PROXY` are supported through
-  [`node-proxy-agent`](https://github.com/TooTallNate/node-proxy-agent).
+- `HTTP_PROXY`, `HTTPS_PROXY`, and `NO_PROXY` are supported through Undici's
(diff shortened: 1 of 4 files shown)


merceyz:
Note that it's experimental and will emit a warning https://github.com/nodejs/undici/blob/532b7b28827afe52ed7b43cbb210ac8425516c28/lib/dispatcher/env-http-proxy-agent.js#L24-L29

aduh95:
Superseded by https://github.com/nodejs/corepack/commit/8167b0d65acc091af464ead019f7485cb512082f

PR #642, opened 2025-02-08
feat: add support for `.corepack.env`

aduh95:
This would allow project author to customize the behavior of Corepack – with the recent incident related to npm registry key rotation, it show how it would be useful to override the built-in values. It could also be used to disable auto pinning at a project level.

It's also a first step towards allowing specifying ranges in the `package.json` (see #634), which has been requested for a long long time (https://github.com/nodejs/corepack/issues/95).

Fixes: https://github.com/nodejs/corepack/issues/628

### DIFF
--- a/README.md
+++ b/README.md
@@ -295,7 +295,7 @@ same major line. Should you need to upgrade to a new major, use an explicit
 
 - `COREPACK_ENV_FILE` can be set to `0` to request Corepack to not attempt to
   load `.corepack.env`; it can be set to a path to specify a different env file.
-  Only keys that starts with `COREPACK_` will be taken into account, not all
+  Only keys that start with `COREPACK_` will be taken into account, not all
   keys that start with `COREPACK_` will be taken into account (
   `COREPACK_ENABLE_DOWNLOAD_PROMPT` and `COREPACK_ENV_FILE` are ignored).
   For Node.js 18.x users, this setting has no effect as that version doesn't
--- a/README.md
+++ b/README.md
@@ -295,9 +295,9 @@ same major line. Should you need to upgrade to a new major, use an explicit
 
 - `COREPACK_ENV_FILE` can be set to `0` to request Corepack to not attempt to
   load `.corepack.env`; it can be set to a path to specify a different env file.
-  Only keys that start with `COREPACK_` will be taken into account, not all
-  keys that start with `COREPACK_` will be taken into account (
-  `COREPACK_ENABLE_DOWNLOAD_PROMPT` and `COREPACK_ENV_FILE` are ignored).
+  Only keys that start with `COREPACK_` and are not in the exception list
+  (`COREPACK_ENABLE_DOWNLOAD_PROMPT` and `COREPACK_ENV_FILE` are ignored)
+  will be taken into account.
   For Node.js 18.x users, this setting has no effect as that version doesn't
   support parsing of `.env` files.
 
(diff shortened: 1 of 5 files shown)


styfle:
I'm not sure about this feature.

Adding a config file sounds like increased complexity for very little benefit.

Do you have examples of when you might use this?

README.md
```diff
@@ -292,6 +293,14 @@ same major line. Should you need to upgrade to a new major, use an explicit
   project. This means that it will always use the system-wide package manager
   regardless of what is being specified in the project's `packageManager` field.
 
+- `COREPACK_ENV_FILE` can be set to `0` to request Corepack to not attempt to
+  load `.corepack.env`; it can be set to a path to specify a different env file.
+  Only keys that starts with `COREPACK_` will be taken into account, not all
+  keys that start with `COREPACK_` will be taken into account (
+  `COREPACK_ENABLE_DOWNLOAD_PROMPT` and `COREPACK_ENV_FILE` are ignored).
```
arcanis, line 300: This would be worth a test, especially for `COREPACK_ENABLE_DOWNLOAD_PROMPT` (assuming you omitted it for security reasons?)

aduh95: Good call (yes, btw I’m happy to add more which we think could be security sensitive)

arcanis (approved):

README.md
```diff
@@ -292,6 +293,14 @@ same major line. Should you need to upgrade to a new major, use an explicit
   project. This means that it will always use the system-wide package manager
   regardless of what is being specified in the project's `packageManager` field.
 
+- `COREPACK_ENV_FILE` can be set to `0` to request Corepack to not attempt to
+  load `.corepack.env`; it can be set to a path to specify a different env file.
+  Only keys that starts with `COREPACK_` will be taken into account, not all
```
MikeMcC399, line 298:
```suggestion
  Only keys that start with `COREPACK_` will be taken into account, not all
```

MikeMcC399: I would suggest to remove "not all  keys that start with `COREPACK_` will be taken into account" and just put "(however `COREPACK_ENABLE_DOWNLOAD_PROMPT` and `COREPACK_ENV_FILE` are ignored)."

The sentence is otherwise a little difficult to read.

MikeMcC399: I understood the meaning to be:

Only keys that start with `COREPACK_` will be taken into account, except `COREPACK_ENABLE_DOWNLOAD_PROMPT` and `COREPACK_ENV_FILE`, which are ignored.

I may have misunderstood though. If that isn't correct, then perhaps you could explain what is missing?

aduh95: Oh right, I guess it even confused me, proving your point. I wrote it this way because I'm only testing for a few, and I wanted to set expectations that supporting all are not a goal

README.md
```diff
@@ -292,6 +293,14 @@ same major line. Should you need to upgrade to a new major, use an explicit
   project. This means that it will always use the system-wide package manager
   regardless of what is being specified in the project's `packageManager` field.
 
+- `COREPACK_ENV_FILE` can be set to `0` to request Corepack to not attempt to
+  load `.corepack.env`; it can be set to a path to specify a different env file.
+  Only keys that start with `COREPACK_` will be taken into account, not all
+  keys that start with `COREPACK_` will be taken into account (
+  `COREPACK_ENABLE_DOWNLOAD_PROMPT` and `COREPACK_ENV_FILE` are ignored).
```
aduh95, line 300:
```suggestion
  Only keys that start with `COREPACK_` and are not in the exception list
  (`COREPACK_ENABLE_DOWNLOAD_PROMPT` and `COREPACK_ENV_FILE` are ignored)
  will be taken into account.
```

Issue #316, opened 2023-10-11
Corepack creates invalid version


rotu:
`corepack use` and `corepack up` add build metadata to the version, resulting in a semver string that doesn't match any released version.

e.g. `"packageManager": "npm@10.2.0+sha256.c362077587b1e782e5aef3dcf85826399ae552ad66b760e2585c4ac11102243f"`

The actual version is simply "10.2.0".

This causes problems for tools that expect to find an existing version: https://github.com/moonrepo/proto/issues/231

Additionally, `corepack` does not tolerate existing build metadata:

e.g. `"packageManager": "npm@9.7.0+some-custom-build"`
```
corepack install
Adding npm@9.7.0+some-custom-build to the cache...
Internal Error: Digest method not supported
```

Related: https://github.com/npm/cli/issues/1479

aduh95:
Why shouldn't Corepack append hash to the version? Surely you understand that it was security concerns that led to that decision, what are your thoughts on that?
As you probably read in the original PR introducing the functionality, the original design did not use a semver compatible syntax. We ended up using a semver compatible one because it's easier to parse, both for Corepack and for other tools who want to support the `packageManager` field.

> Additionally, `corepack` does not tolerate existing build metadata

If that's a problem for someone, PRs welcome, but I doubt that there's a use case for specifying metadata except a hash.

rotu:
>Why shouldn't Corepack append hash to the version? I can see you link to the original PR introducing the functionality, you have probably read that it was security concerns that led to that decision, what are your thoughts on that?

There are a couple of reasons:

1. It makes the `packageManager` field no longer human-friendly to edit, since it now has a human-readable aspect and a machine-generated aspect.
2. The version is at the discretion of the author of the package; this hash is an artifact of this tool, not part of the package version.
3. The hash is not a property of the code; it's a property of the tarball. It depends not just on the version but on the "resolved" url from `package-lock.json` (which has no analog in corepack and will be a problem for private repositories).

Yes, I think it makes sense to hash the package manager. It doesn't make sense to put it in the version string.

aduh95:
Then don’t push the hash / edit it out, no? If you want human editable, it seems only faire that you have to edit it manually, and it’s completely optional. But I have a hard time seeing why you would not chose to give up the security of the hash. Or do you have a suggestion to store the hash somewhere else?

rotu:
[...]

PR #643, opened 2025-02-08
feat: add limited support for `devEngines`

aduh95:
This implementation of `devEngines` support in this PR is limited to `packageManager` only, only one `packageManager` can be specified, ~the `onFail` field is ignored~ EDIT: I've added support for the `onFail` field, supporting the same values as npm.

Fixes: https://github.com/nodejs/corepack/issues/567

### DIFF
--- a/README.md
+++ b/README.md
@@ -113,6 +113,19 @@ use in the archive).
 }
 ```
 
+#### `devEngines.packageManager`
+
+When a `devEngines.packageManager` field is defined, and is an object containing
+a `"name"` field (can also optionally contain `version` and `onFail` fields),
+Corepack will use it to validate you're using a compatible package manager.
+
+Depending on the value of `devEngines.packageManager.onFail`:
+
+- if set to `ignore`, Corepack won't print any warning or error.
+- if unset or set to `error`, Corepack will throw an error in case of a mismatch.
+- if set to `warn` or some other value, Corepack will print a warning in case
+  of mismatch.
+
 ## Known Good Releases
 
 When running Corepack within projects that don't list a supported package
@@ -246,6 +259,7 @@ it.
 
 Unlike `corepack use` this command doesn't take a package manager name nor a
 version range, as it will always select the latest available version from the
+range specified in `devEngines.packageManager.version`, or fallback to the
 same major line. Should you need to upgrade to a new major, use an explicit
 `corepack use {name}@latest` call (or simply `corepack use {name}`).
 
--- a/README.md
+++ b/README.md
@@ -126,6 +126,22 @@ Depending on the value of `devEngines.packageManager.onFail`:
 - if set to `warn` or some other value, Corepack will print a warning in case
   of mismatch.
 
+If the top-level `packageManager` field is missing, Corepack will use the
+package manager defined in `devEngines.packageManager` – in which case you must
+provide a specific version in `devEngines.packageManager.version`, ideally with
+a hash, as explained in the previous section:
+
+```json
+{
+  "devEngines":{
+    "packageManager": {
+      "name": "yarn",
+      "version": "3.2.3+sha224.953c8233f7a92884eee2de69a1b92d1f2ec1655e66d08071ba9a02fa"
+    }
+  }
+}
+```
+
 ## Known Good Releases
 
 When running Corepack within projects that don't list a supported package
--- a/sources/commands/Base.ts
+++ b/sources/commands/Base.ts
@@ -16,7 +16,7 @@ export abstract class BaseCommand extends Command<Context> {
           throw new UsageError(`Couldn't find a project in the local directory - please explicit the package manager to pack, or run this command from a valid project`);
 
         case `NoSpec`:
-          throw new UsageError(`The local project doesn't feature a 'packageManager' field nor a 'devEngines.packageManager' field - please explicit the package manager to pack, or update the manifest to reference it`);
+          throw new UsageError(`The local project doesn't feature a 'packageManager' field nor a 'devEngines.packageManager' field - please specify the package manager to pack, or update the manifest to reference it`);
 
         default: {
           return [lookup.range ?? lookup.spec];
--- a/sources/commands/Base.ts
+++ b/sources/commands/Base.ts
@@ -19,7 +19,7 @@ export abstract class BaseCommand extends Command<Context> {
           throw new UsageError(`The local project doesn't feature a 'packageManager' field - please explicit the package manager to pack, or update the manifest to reference it`);
 
         default: {
-          return [lookup.spec];
+          return [lookup.range ?? lookup.spec];
         }
       }
     }
--- a/sources/commands/Base.ts
+++ b/sources/commands/Base.ts
@@ -16,7 +16,7 @@ export abstract class BaseCommand extends Command<Context> {
           throw new UsageError(`Couldn't find a project in the local directory - please explicit the package manager to pack, or run this command from a valid project`);
 
         case `NoSpec`:
-          throw new UsageError(`The local project doesn't feature a 'packageManager' field - please explicit the package manager to pack, or update the manifest to reference it`);
+          throw new UsageError(`The local project doesn't feature a 'packageManager' field nor a 'devEngines.packageManager' field - please explicit the package manager to pack, or update the manifest to reference it`);
 
         default: {
           return [lookup.range ?? lookup.spec];
--- a/sources/specUtils.ts
+++ b/sources/specUtils.ts
@@ -62,6 +62,18 @@ type CorepackPackageJSON = {
 interface DevEngineDependency {
   name: string;
   version: string;
+  onFail?: 'ignore' | 'warn' | 'error';
+}
+function warnOrThrow(errorMessage: string, onFail?: DevEngineDependency['onFail']) {
+  switch (onFail) {
+    case `ignore`:
+      break;
+    case `error`:
+    case undefined:
+      throw new UsageError(errorMessage);
+    default:
+      console.warn(`! Corepack validation warning: ${errorMessage}`);
+  }
 }
 function parsePackageJSON(packageJSONContent: CorepackPackageJSON) {
   if (packageJSONContent.devEngines?.packageManager != null) {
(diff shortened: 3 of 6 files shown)


sources/specUtils.ts
```diff
@@ -52,6 +53,46 @@ export function parseSpec(raw: unknown, source: string, {enforceExactVersion = t
...
+interface DevEngineDependency {
+  name: string;
+  version: string;
+}
+function parsePackageJSON(packageJSONContent: CorepackPackageJSON) {
+  if (packageJSONContent.devEngines?.packageManager) {
+    const {packageManager} = packageJSONContent.devEngines;
+
+    if (Array.isArray(packageManager))
+      throw new UsageError(`Providing several package managers is currently not supported`);
```
GeoffreyBooth, line 70: Perhaps an array of length 1 should be permitted?

sources/specUtils.ts
```diff
@@ -52,6 +53,46 @@ export function parseSpec(raw: unknown, source: string, {enforceExactVersion = t
...
+
+    const {version} = packageManager;
+    if (!version)
+      throw new UsageError(`Providing no version nor ranger for package manager is currently not supported`);
```
GeoffreyBooth, line 74: Spelling, and also the double negative is confusing.

```suggestion
      throw new UsageError(`Package manager version or version range is required`);
```

tests/main.test.ts
```diff
@@ -231,6 +258,82 @@ it(`should ignore the packageManager field when found within a node_modules vend
   });
 });
 
+it(`should use hash from "packageManager" even when "devEngines" defines a different one`, async () => {
```
GeoffreyBooth, line 261: Shouldn’t this error?

sources/specUtils.ts
```diff
@@ -52,6 +53,46 @@ export function parseSpec(raw: unknown, source: string, {enforceExactVersion = t
...
+
+    debugUtils.log(`devEngines defines that ${packageManager.name}@${version} is the local package manager`);
+
+    const {packageManager: pm} = packageJSONContent;
```
GeoffreyBooth, line 78: So is the `packageManager` field still required, even if the user specifies `devEngines`?

tests/main.test.ts
```diff
@@ -16,16 +16,33 @@ beforeEach(async () => {
...
-it(`should refuse to download a package manager if the hash doesn't match`, async () => {
-  await xfs.mktempPromise(async cwd => {
-    await xfs.writeJsonPromise(ppath.join(cwd, `package.json` as Filename), {
-      packageManager: `yarn@1.22.4+sha1.deadbeef`,
+describe(`should refuse to download a package manager if the hash doesn't match`, () => {
+  it(`the one defined in "devEngines.packageManager" field`, async () => {
+    await xfs.mktempPromise(async cwd => {
+      await xfs.writeJsonPromise(ppath.join(cwd, `package.json` as Filename), {
+        devEngines: {
+          packageManager: {name: `yarn`, version: `1.22.4+sha1.deadbeef`},
```
GeoffreyBooth, line 24: Does adding a hash to the `version` value break other tools that support `devEngines`? For example, when pnpm 10.2.1 reads a `devEngines` with a value of:

```json
"devEngines": {
  "packageManager": {
    "name": "pnpm",
    "version": "10.2.1+sha1.48adf39a4ab751eda7b73b99447d1f0b6d227e02"
  }
}
```

does it still pass pnpm’s validation that the correct package manager version is specified?

sources/specUtils.ts
aduh95: IMO that can be done as a follow up when adding support for arrays. I’d prefer to keep the implementation simpler for the initial PR

tests/main.test.ts
aduh95: I don’t think so, according to semver spec, two versions with same number and different build metadata should refer to the same release. In https://github.com/nodejs/corepack/issues/316 someone requested to have the possibility to put build metadata that’s not a hash, so I figured we could use devEngines for that

sources/specUtils.ts
aduh95: That’s correct

tests/main.test.ts
aduh95: According to semver spec, it shouldn’t, but anyway that’s off topic for this PR, if it’s not compatible with the package manager they’re using, users won’t supply it

sources/specUtils.ts
GeoffreyBooth: Do you mind adding an explanation in the PR description for what you think the eventual complete support looks like? So like that Corepack will eventually support arrays; will it support `onFail` too? Etc.

tests/main.test.ts
GeoffreyBooth: > two versions with same number and different build metadata

So like one has a sha1 hash and the other a sha512 hash, something like that? I was thinking of two hashes of the same type but different values, and I was wondering how that could ever be something other than an error (since only one could be correct).

aduh95: I'm using that part of the semver spec as reference:

> Build metadata MUST be ignored when determining version precedence. Thus two versions that differ only in the build metadata, have the same precedence. Examples: 1.0.0-alpha+001, 1.0.0+20130313144700, 1.0.0-beta+exp.sha.5114f85, 1.0.0+21AF26D3----117B344092BD.

My understanding is that it doesn't matter whether the build metadata refer to a hash or not, as long as it has the same version number, it refers to the same version, and therefor should pass validation – as long as the hash in `packageManager` corresponds to the download archive.

sources/specUtils.ts
aduh95: I think the PR description is describing the current PR, I don t think it should describe potential future PRs.

styfle:
> the onFail field is ignored

What value of onFail is used?

sources/specUtils.ts
GeoffreyBooth: Without an explanation, it's unclear how much of the spec Corepack intends to support. And if the support lands piecemeal and is released that way, there might be weird gaps where particular versions of Corepack error on things (like arrays) that other versions don't.

If you flag it or release it all together somehow, I think that's fine; but otherwise maybe just implement all of the spec that you think makes sense for Corepack to implement? The devEngines spec isn't that large.

aduh95: I've changed the throw into a warning

arcanis (approved):

styfle (commented):
How about updating the readme?

In particular, mentioning the onFail behavior and also which takes precedence when both packageManager and devEngines is defined?

GeoffreyBooth (approved):

README.md
```diff
@@ -113,6 +113,35 @@ use in the archive).
...
+Corepack will use it to validate you're using a compatible package manager.
+
+Depending on the value of `devEngines.packageManager.onFail`:
+
+- if set to `ignore`, Corepack won't print any warning or error.
+- if unset or set to `error`, Corepack will throw an error in case of a mismatch.
+- if set to `warn` or some other value, Corepack will print a warning in case
+  of mismatch.
+
+If the top-level `packageManager` field is missing, Corepack will use the
```
styfle, line 129: This should be the first sentence of this section.

Because it sounds like this `devEngines.packageManager` is ignored when there is a top level `packageManager`, right?

README.md
```diff
@@ -246,6 +275,7 @@ it.
 
 Unlike `corepack use` this command doesn't take a package manager name nor a
 version range, as it will always select the latest available version from the
+range specified in `devEngines.packageManager.version`, or fallback to the
```
styfle, line 278: What about top-level `packageManager`?

sources/specUtils.ts
```diff
@@ -52,6 +54,70 @@ export function parseSpec(raw: unknown, source: string, {enforceExactVersion = t
...
+      break;
+    case `error`:
+    case undefined:
+      throw new UsageError(errorMessage);
+    default:
+      console.warn(`! Corepack validation warning: ${errorMessage}`);
+  }
+}
+function parsePackageJSON(packageJSONContent: CorepackPackageJSON) {
+  if (packageJSONContent.devEngines?.packageManager != null) {
```
styfle, line 79: Now I'm confused because the code seems to prefer `devEngines.packageManager` over the top-level `packageManager` but the README says the opposite.

README.md
aduh95: No it's never ignored. As said in the first paragraph of this section, if it is defined to something Corepack recognises, it will throw/warn/do nothing (depending of the value of `onFail`) if the user is trying to use an incompatible package manager (i.e. if the one defined in `package.json#packageManager` does not match the one defined in `package.json#devEngines.packageManager`)

README.md
aduh95: That's what `or fallback to the same major line` is for

sources/specUtils.ts
aduh95: You misreading the code, we're always returning the top-level field when it's defined

styfle (approved):

sources/commands/Base.ts
```diff
@@ -19,7 +19,7 @@ export abstract class BaseCommand extends Command<Context> {
           throw new UsageError(`The local project doesn't feature a 'packageManager' field - please explicit the package manager to pack, or update the manifest to reference it`);
```
aduh95, line 19:
```suggestion
          throw new UsageError(`The local project doesn't feature a 'packageManager' field nor a 'devEngines.packageManager' field - please explicit the package manager to pack, or update the manifest to reference it`);
```

sources/commands/Base.ts
```diff
@@ -16,10 +16,10 @@ export abstract class BaseCommand extends Command<Context> {
           throw new UsageError(`Couldn't find a project in the local directory - please explicit the package manager to pack, or run this command from a valid project`);
 
         case `NoSpec`:
-          throw new UsageError(`The local project doesn't feature a 'packageManager' field - please explicit the package manager to pack, or update the manifest to reference it`);
+          throw new UsageError(`The local project doesn't feature a 'packageManager' field nor a 'devEngines.packageManager' field - please explicit the package manager to pack, or update the manifest to reference it`);
```
MikeMcC399, line 19: "explicit" is not a verb. I suggest to replace with "specify" also in line 16

aduh95: It should be
```suggestion
          throw new UsageError(`The local project doesn't feature a 'packageManager' field nor a 'devEngines.packageManager' field - please specify the package manager to pack, or update the manifest to reference it`);
```

sources/commands/Base.ts
```diff
@@ -16,10 +16,10 @@ export abstract class BaseCommand extends Command<Context> {
           throw new UsageError(`Couldn't find a project in the local directory - please explicit the package manager to pack, or run this command from a valid project`);
```
MikeMcC399, line 16:
```suggestion
          throw new UsageError(`Couldn't find a project in the local directory - please specify the package manager to pack, or run this command from a valid project`);
```

aduh95: Would you like to send a separate PR? That feels a bit unrelated

MikeMcC399: Sure. I hadn't really checked where this came from.

aduh95:
Blocked on https://github.com/nodejs/corepack/pull/663

merdopo66 (approved):

merdopo66 (approved):

tests/main.test.ts
```diff
@@ -237,6 +268,298 @@ it(`should ignore the packageManager field when found within a node_modules vend
...
+
+    await expect(runCli(cwd, [`yarn`, `--version`])).resolves.toMatchObject({
+      exitCode: 1,
+      stderr: expect.stringContaining(`Mismatch hashes. Expected 11111, got`),
+      stdout: ``,
+    });
+  });
+});
+
+describe(`should accept range in devEngines only if a specific version is provided`, () => {
```
susnux, line 380: This behavior breaks existing projects which only define a range in `devEngines` (e.g. to make developers all use same major line of the package manager).
Normally that would be somewhat ok as its only your project and you know what you are doing, but this affects a lot of different tools like **dependabot** that are now broken for projects that not use `corepack` themself.

Issue #612, opened 2025-01-27
Newly published versions of package managers distributed from npm cannot be installed due to key id mismatch


sapphi-red:
This is probably related to #611.

When I try to install pnpm 10.1.0, I get the following error:
```
C:\Program Files\nodejs\node_modules\corepack\dist\lib\corepack.cjs:21535
  if (key == null || signature == null) throw new Error(`Cannot find matching keyid: ${JSON.stringify({ signatures, keys })}`);
                                              ^

Error: Cannot find matching keyid: {"signatures":[{"keyid":"SHA256:DhQ8wR5APBvFHLF/+Tc+AYvPOdTpcIDqOhxsBHRwC7U","sig":"MEUCIQDlkgmNyZjT7KUY8AO6jH7Gs3fyiXG8nbTnuLbd8fOS2AIgXyJ6SaYhumMFzUYQAZPJGhsnlaD5N0X2MZsbG+eS/Xo="}],"keys":[{"expires":null,"keyid":"SHA256:jl3bwswu80PjjokCgh0o2w5c2U4LhQAE57gj9cz1kzA","keytype":"ecdsa-sha2-nistp256","scheme":"ecdsa-sha2-nistp256","key":"MFkwEwYHKoZIzj0CAQYIKoZIzj0DAQcDQgAE1Olb3zMAFFxXKHiIkQO5cJ3Yhl5i6UPp+IhuteBJbuHcA5UogKo0EWtlWwW6KSaKoTNEYL7JlCQiVnkhBktUgg=="}]}
    at verifySignature (C:\Program Files\nodejs\node_modules\corepack\dist\lib\corepack.cjs:21535:47)
    at installVersion (C:\Program Files\nodejs\node_modules\corepack\dist\lib\corepack.cjs:21882:7)
    at process.processTicksAndRejections (node:internal/process/task_queues:105:5)
    at async Engine.ensurePackageManager (C:\Program Files\nodejs\node_modules\corepack\dist\lib\corepack.cjs:22316:32)
    at async Engine.executePackageManagerRequest (C:\Program Files\nodejs\node_modules\corepack\dist\lib\corepack.cjs:22416:25)
    at async Object.runMain (C:\Program Files\nodejs\node_modules\corepack\dist\lib\corepack.cjs:23102:5)        

Node.js v22.13.0
```

I checked https://registry.npmjs.org/pnpm/10.1.0 and it contains
```json
{
  "signatures": [
    {
[...]

PR #647, opened 2025-02-11
Get NPM signing keys from @sigstore/tuf

### LINKED ISSUE #616: Saveguards for keyid mismatch
- Issue https://github.com/nodejs/corepack/issues/612 records a disruption to the corepack service after npm registry keys were updated.

- How can this be prevented in future? I'm not familiar with how the keys and systems for using them work, so this is a question for the experts.

---
In https://github.com/nodejs/corepack/issues/612#issuecomment-2616588603 @wraithgar wrote
> https://registry.npmjs.org/-/npm/v1/keys should now be returning two keys. During key rotation the old one was omitted by mistake and has since been re-added.

Even if a mistake hadn't been made, would there still have been an issue needing keys urgently updating in Corepack?

Does Corepack need to know in advanced if npm registry keys on the endpoint https://registry.npmjs.org/-/npm/v1/keys (see https://docs.npmjs.com/about-registry-signatures) are going to be changed?

If npm registry keys are changed, is there any advanced notice from the npm registry of a plan to change?

Is there any notification that npm registry keys have changed, at the time of change or does the endpoint need to be polled by Corepack regularly?

Can Corepack ensure that the version bundled in Node.js will always have a set of keys which can be used without error?

geigerzaehler:
Instead of hardcoding NPM signing keys for verification we get them from sigstore’s TUF repository. This is in line with how npm implements signature verification.

Fixes #616, #612

---

The implementation is ready to review, @aduh95. The only thing missing is a test for failed signature verification that does use `COREPACK_INTEGRITY_KEYS`. If you can think of more tests, let me know. I’m also not sure how the test setup works with regards to recording http requests.

- [ ] Test that signature verification fails if providing an invalid signature for TUF keys.


geigerzaehler:
This change also does not check for key expiry. This requires some more investigation and I’ll tackle this in a follow-up change.

sources/npmRegistryUtils.ts
```diff
@@ -32,30 +34,58 @@ export async function fetchAsJson(packageName: string, version?: string) {
...
-
+  const keys = await getVerificationKeys();
   const key = keys.find(({keyid}) => signatures.some(s => s.keyid === keyid));
-  const signature = signatures.find(({keyid}) => keyid === key?.keyid);
+  if (key == null)
+    throw new Error(`Cannot find key to verify signature. signature keys: ${signatures.map(s => s.keyid)}, verification keys: ${keys.map(k => k.keyid)}`);
 
-  if (key == null || signature == null) throw new Error(`Cannot find matching keyid: ${JSON.stringify({signatures, keys})}`);
+  const signature = signatures.find(({keyid}) => keyid === key.keyid);
+  assert(signature); // If `key` is defined there is a matching signature
```
OR13, line 77: this comment could be clearer

aduh95:
This makes the bundle size go from 936.5kb to 2.4mb, do we actually need all that code?

sources/npmRegistryUtils.ts
```diff
@@ -32,30 +34,58 @@ export async function fetchAsJson(packageName: string, version?: string) {
...
+    keys = keysFromEnv.npm.map(k => ({keyid: k.keyid, key: k.key}));
+    debugUtils.log(`Using COREPACK_INTEGRITY_KEYS to verify signatures: ${keys.map(k => k.keyid).join(`, `)}`);
+  } else {
+    // This follows the implementation for npm.
+    // See https://github.com/npm/cli/blob/3a80a7b7d168c23b5e297cba7b47ba5b9875934d/lib/utils/verify-signatures.js#L174
+    // We only support sigstore for NPM. For other registries
+    // COREPACK_INTEGRITY_KEYS can be used.
+    const sigstoreTufClient = await sigstoreTuf.initTUF();
+    const keysRaw = await sigstoreTufClient.getTarget(`registry.npmjs.org/keys.json`);
+    // The format of the key file is undocumented, unfortunately. `rawBytes` is
```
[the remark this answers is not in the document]
aduh95: > Not sure if there is ever plan to support other signing schemes beyond "ecdsa-sha2-nistp256".

I think the recent incident has taught us that it's unwise to assume the current npm setup won't change on us without any warning signs, so let's assume that yes, other key types are coming

aduh95:
Maybe we could try patching `tuf-js` to use `undici` instead of `make-fetch-happen` (https://github.com/theupdateframework/tuf-js/blob/0303dab1e09cd2c58b9707c4c2b57566c7df349e/packages/client/src/fetcher.ts#L3)

config.json
```diff
@@ -161,23 +161,5 @@
...
-        "key": "MFkwEwYHKoZIzj0CAQYIKoZIzj0DAQcDQgAE1Olb3zMAFFxXKHiIkQO5cJ3Yhl5i6UPp+IhuteBJbuHcA5UogKo0EWtlWwW6KSaKoTNEYL7JlCQiVnkhBktUgg=="
-      },
-      {
-        "expires": null,
-        "keyid": "SHA256:DhQ8wR5APBvFHLF/+Tc+AYvPOdTpcIDqOhxsBHRwC7U",
-        "keytype": "ecdsa-sha2-nistp256",
-        "scheme": "ecdsa-sha2-nistp256",
-        "key": "MFkwEwYHKoZIzj0CAQYIKoZIzj0DAQcDQgAEY6Ya7W++7aUPzvMTrezH6Ycx3c+HOKYCcNGybJZSCJq/fd7Qa8uuAKtdIkUQtQiEKERhAmE5lMMJhP8OkDOa2g=="
-      }
-    ]
```
aduh95, line 181: Shouldn't we keep this as a fallback?

sources/npmRegistryUtils.ts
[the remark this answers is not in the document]
geigerzaehler: I don’t think we need to care about "keytype" and "scheme" and the code will work with other schemes. The assumption here is that the `publicKey.rawBytes` is the content of a PUBLIC KEY PEM string. That means  that the key [is encoded with SPKI DER](https://datatracker.ietf.org/doc/html/rfc7468#section-13) which identifies the key type and consequently the signing scheme. `createPublicKey` from `node:crypto` will then correctly create the correct key and the verifier will use the correct scheme.

config.json
geigerzaehler: Not sure. This would only make a difference if fall back to them if we can’t fetch the sigstore keys. I’m a bit worried that this redundancy will obscure problems that, in the end, lead to the same problem as we have now. 

This is because the fallback mechanism will break if a new key is added. This will not be announced and it will stay broken unless we maintain it. But we may not notice it is broken because the fallback is not triggered. So it looks like forever maintenance burden that doesn’t actually add that much.

aduh95: Well we would have tests for it, that's how we detected it this time around, I expect it would be different next time

geigerzaehler:
> EDIT: I've tried doing that, if we're fine getting rid of the `retry: 2` default option, with the following patch we're down to a 1.2MiB bundle:

That’s great. Would you like me to add this as a yarn patch? Or do you want to get this into `tuf-js`?

aduh95:
> > EDIT: I've tried doing that, if we're fine getting rid of the `retry: 2` default option, with the following patch we're down to a 1.2MiB bundle:
> 
> That’s great. Would you like me to add this as a yarn patch? Or do you want to get this into `tuf-js`?

I think we should add it as a Yarn patch. I can certainly send a PR their way, but given it would be a major change for them, I don't think it would make sense for us to wait on them.

config.json
geigerzaehler: Ok. Can you elaborate how we should do the fallback? Should we use the hard-coded keys when the TUF requests fails? Do we want to distinguish between different failures (i.e. network vs verification, etc.)? Or should we include the hard-coded keys in case they are missing from the returned from the TUF client?

aduh95: IMO the trust hierarchy should be:

1. env (`COREPACK_INTEGRITY_KEYS`)
2. TUF
3. `config.json`

We should use the most trusted key set at our disposal, and hard fail only if signature doesn't match / the key is unknown (or known to be expired, but support for key expiry should be a separate PR IMO).
So a network issue should simply fallback to `config.json` IMO, and not be reported to the user (or rather only with `debugUtils.log`).

geigerzaehler: I gave this a bit more thought and I believe the extra keys will decrease the overall security by increasing the attack surface.

1. If an attacker can compromise *either* of the keys they can trick into accepting forged signatures.
2. If a key is compromised, it has to be revoked from both TUF *and* `config.json`.

The problem is that there is no trust hierarchy, either we accept signatures from a key or we don’t. Because trust is so important, my recommendation would be to limit the points of trust as much as possible and not include hard-coded keys. 

To elaborate a bit on 1: In practice this probably does not make a huge difference because if somebody can compromise the hard-coded key in the source they can most likely also compromise the dependency code in the bundle.

From a usability perspective relying just on TUF is not a huge issue either, I believe. If it is a transient issue (i.e. network, service availability) we can provide a meaningful error an users can always retry later. This happens all the time and any user of `npm` will be similarly affected. Any other issue is either an attack or a change in the signature mechanism. In the first case, we should fail. In the second case there will be a migration period, again `npm` also has to phase out the mechanism.

In any case, the user can always workaround the issue by setting `COREPACK_INTEGRITY_KEYS`.

What do you think? Does this make sense?

aduh95: > 2\. If a key is compromised, it has to be revoked from both TUF _and_ `config.json`.

Why would that be the case? It can be revoked only on TUF since it's up the trust hierarchy (that is, until Corepack is updated and `config.json` contains up-to-date info)

sources/npmRegistryUtils.ts
```diff
@@ -32,30 +34,58 @@ export async function fetchAsJson(packageName: string, version?: string) {
...
 }
 
-export function verifySignature({signatures, integrity, packageName, version}: {
+const getVerificationKeys = once(async (): Promise<Array<{ keyid: string, key: KeyObject }>> => {
+  let keys: Array<{ keyid: string, key: string }>;
+  if (process.env.COREPACK_INTEGRITY_KEYS) {
+    // We use the format of the `GET /-/npm/v1/keys` endpoint with `npm` instead
+    // of `keys` as the wrapping key.
+    const keysFromEnv = JSON.parse(process.env.COREPACK_INTEGRITY_KEYS) as { npm: Array<{ keyid: string, key: string }> };
+    keys = keysFromEnv.npm.map(k => ({keyid: k.keyid, key: k.key}));
```
aduh95, line 43: That seems undesirable, we should use whatever value was passed by the user
```suggestion
```

sources/npmRegistryUtils.ts
```diff
@@ -32,30 +34,58 @@ export async function fetchAsJson(packageName: string, version?: string) {
...
+    debugUtils.log(`Using COREPACK_INTEGRITY_KEYS to verify signatures: ${keys.map(k => k.keyid).join(`, `)}`);
+  } else {
+    // This follows the implementation for npm.
+    // See https://github.com/npm/cli/blob/3a80a7b7d168c23b5e297cba7b47ba5b9875934d/lib/utils/verify-signatures.js#L174
+    // We only support sigstore for NPM. For other registries
+    // COREPACK_INTEGRITY_KEYS can be used.
+    const sigstoreTufClient = await sigstoreTuf.initTUF();
```
aduh95, line 50: We probably want to provide a cachePath, as npm is doing https://github.com/npm/cli/blob/3a80a7b7d168c23b5e297cba7b47ba5b9875934d/workspaces/config/lib/definitions/definitions.js#L315
```suggestion
    const sigstoreTufClient = await sigstoreTuf.initTUF({ cachePath: path.join(folderUtils.getCorepackHomeFolder(), `_tuf`) });
```

sources/npmRegistryUtils.ts
OR13: I would tend to remove the comment.

The comment is really about the interaction between these 3 lines:

```
// find a key that appears to produce a signature
const key = keys.find(({keyid}) => signatures.some(s => s.keyid === keyid));
// find the first signature produced by the key
const signature = signatures.find(({keyid}) => keyid === key.keyid);
// ensure there exists a signature to verify
assert(signature);
```

sources/npmRegistryUtils.ts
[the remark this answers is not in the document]
OR13: You would hope that a key type implied a single signature scheme.

```
NPM_KEY_FORMAT="MFkwEwYHKoZIzj0CAQYIKoZIzj0DAQcDQgAE1Olb3zMAFFxXKHiIkQO5cJ3Yhl5i6UPp+IhuteBJbuHcA5UogKo0EWtlWwW6KSaKoTNEYL7JlCQiVnkhBktUgg=="
PEM_KEY="-----BEGIN PUBLIC KEY-----\n$NPM_KEY_FORMAT\n-----END PUBLIC KEY-----"
echo $PEM_KEY > pub.key
openssl ec -inform PEM -pubin -in pub.key -text -noout
```

```
read EC key
Public-Key: (256 bit)
pub:
    04:d4:e9:5b:df:33:00:14:5c:57:28:78:88:91:03:
    b9:70:9d:d8:86:5e:62:e9:43:e9:f8:88:6e:b5:e0:
    49:6e:e1:dc:03:95:28:80:aa:34:11:6b:65:5b:05:
    ba:29:26:8a:a1:33:44:60:be:c9:94:24:22:56:79:
    21:06:4b:54:82
ASN1 OID: prime256v1
NIST CURVE: P-256
```

Did I miss something? 

There is no signature algorithm or hash function implied here.

The part above where you construct the verifier is where you need to accept explicitly the signature scheme, and any parameters of it needed for interop, such as hash function, which for these npm signatures is "ecdsa-sha2-nistp256".

Consider the case where the hash function changes, if you don't look at "keytype" and "signature" how would you know to choose between sha256 and sha512 ?

The code will work with other schemes, assuming you construct the verifier with the correct hash function for the scheme.

Ed25519 or ML-DSA are going to end up with different verification logic, since they are not parameterized by hash function.

config.json
aduh95: Using a union would be a plain bad idea, it is not at all my suggestion. My suggestion is to fallback on the hard coded keys when TUF is not available – if TUF ever goes down, we shouldn't have to make emergency releases of Corepack; let's not replace a single point of failure with another single point of failure.

sources/npmRegistryUtils.ts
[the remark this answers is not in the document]
geigerzaehler: You’re totally right @OR13. `createPublicKey` works with different key types but we also need information about the scheme to know how to verify the signature.

I would, however, prefer to tackle support for other schemes in a separate PR to limit the scope and get this through faster. The implementation is not straight-forward, I can’t find any documentation on this and it is not implemented in the `npm` CLI as of now.

> I think the recent incident has taught us that it's unwise to assume the current npm setup won't change on us without any warning signs, so let's assume that yes, other key types are coming

I agree that this should be handled in the future. But I think the situation is a bit different and not as problematic. First, adding a key with a different scheme will not break verification—we’ll just use an old key. Second, if a scheme is phased we’ll have advanced notice since the `npm` CLI only supports the current scheme. So they will have to deprecate the current version of `npm` before the scheme can be removed.

config.json
geigerzaehler: Ok. I have to say that I disagree with this approach since it weakens security and you haven’t addressed the concerns about possible attacks. But I then again, it is not less secure than the current system. So I’ll defer the decision to you and will implement it as suggested.

geigerzaehler:
I’ve implemented the test and addressed all the feedback. This is ready for review now

.yarn/patches/tuf-js-npm-3.0.1-9135d15fbd.patch
```diff
@@ -0,0 +1,31 @@
...
+ const tmpfile_1 = require("./utils/tmpfile");
+@@ -61,14 +61,13 @@ class DefaultFetcher extends BaseFetcher {
+     }
+     async fetch(url) {
+         log('GET %s', url);
+-        const response = await (0, make_fetch_happen_1.default)(url, {
+-            timeout: this.timeout,
+-            retry: this.retry,
++        const response = await globalThis.fetch(url, {
++            timeout: this.timeout && AbortSignal.timeout(this.timeout)
```
aduh95, line 22:
```suggestion
+            signal: this.timeout && AbortSignal.timeout(this.timeout)
```

aduh95:
As pointed out in https://github.com/theupdateframework/tuf-js/pull/849#pullrequestreview-2613597859, we should think about proxy support in our patch (ideally using the same one as Corepack does)

sources/npmRegistryUtils.ts
```diff
@@ -32,30 +37,101 @@ export async function fetchAsJson(packageName: string, version?: string) {
...
+
+  if (process.env.COREPACK_INTEGRITY_KEYS) {
+    // We use the format of the `GET /-/npm/v1/keys` endpoint with `npm` instead
+    // of `keys` as the wrapping key.
+    const keysFromEnv = JSON.parse(process.env.COREPACK_INTEGRITY_KEYS) as {npm: Array<{keyid: string, key: string}>};
+    keys = keysFromEnv.npm;
+    debugUtils.log(`Using COREPACK_INTEGRITY_KEYS to verify signatures: ${keys.map(k => k.keyid).join(`, `)}`);
+    return keys.map(k => ({
+      keyid: k.keyid,
+      key: crypto.createPublicKey(`-----BEGIN PUBLIC KEY-----\n${k.key}\n-----END PUBLIC KEY-----`,
```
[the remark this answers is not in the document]
geigerzaehler: Makes sense. Fixed in https://github.com/nodejs/corepack/pull/647/commits/db1001110c05c18e4501c8a171214f6896db5d39

.yarn/patches/tuf-js-npm-3.0.1-9135d15fbd.patch
geigerzaehler: Fixed in https://github.com/nodejs/corepack/pull/647/commits/db1001110c05c18e4501c8a171214f6896db5d39

geigerzaehler:
> As pointed out in [theupdateframework/tuf-js#849 (review)](https://github.com/theupdateframework/tuf-js/pull/849#pullrequestreview-2613597859), we should think about proxy support in our patch (ideally using the same one as Corepack does)

Addressed in https://github.com/nodejs/corepack/pull/647/commits/3a2115e62501e7e7f8829e5e823fc4063c5fd53a

README.md
```diff
@@ -316,8 +316,17 @@ same major line. Should you need to upgrade to a new major, use an explicit
   [`proxy-from-env`](https://github.com/Rob--W/proxy-from-env).
 
 - `COREPACK_INTEGRITY_KEYS` can be set to an empty string or `0` to
-  instruct Corepack to skip integrity checks, or to a JSON string containing
-  custom keys.
+  instruct Corepack to skip signature verification, or to a JSON string
+  containing custom keys. The format based on the response of the
+  `GET /-/npm/v1/keys` endpoint of npm registry under the `npm` key. That is,
```
MikeMcC399, line 321:
```suggestion
  containing custom keys. The format is based on the response of the
  `GET /-/npm/v1/keys` npm registry endpoint under the `npm` key. That is,
```

geigerzaehler:
It looks like we can’t just use replayed responses for TUF request because `@sigstore/tuf` checks timestamps and those in the recorded responses become stale and the client [throws an error](https://github.com/nodejs/corepack/actions/runs/13494613952/job/37699006114?pr=647). I’m leaning towards not recording TUF requests and rely on network requests in the tests. Alternatively, we could mock the current date. This would be a bit more cumbersome to deal with when requests are re-recorded and would depend on https://github.com/nodejs/corepack/pull/654. WDYT @aduh95?

OR13:
Replay with mocked current time seems most aligned with the intention of the tests.

cupofjoakim:
Is this work ongoing or has it stalled?

geigerzaehler:
I’m working on rebasing this and fixing the last issues. I hope to get this done next week.

geigerzaehler:
I’ve updated this PR and everything should work now. PTAL @aduh95

tests/_registryServer.mjs
```diff
@@ -7,6 +7,14 @@ import {gzipSync}                                    from 'node:zlib';
 let privateKey, keyid;
 
 switch (process.env.TEST_INTEGRITY) {
+  case `invalid_npm_signature`: {
+    // Claim to use a known NPM signing key but provide an invalid signature
```
OR13, line 11:
```suggestion
    // Claim to use a known NPM signing key but provide a signature from a different key
```

tests/recordRequests.js
```diff
@@ -41,11 +42,20 @@ function getRequestHash(input, init) {
...
+
 if (process.env.NOCK_ENV === `record`) {
   const insertNockStatement = db.prepare(`INSERT OR REPLACE INTO nocks (hash, body, headers, status) VALUES (?, ?, jsonb(?), ?)`);
-  const realFetch = globalThis.fetch;
+
   globalThis.fetch = async (input, init) => {
-    const response = await realFetch(input, init);
+    if (passthroughUrls.some(passThrough => input.toString().startsWith(passThrough)))
+      return originalFetch(input, init);
+    if (process.env.BLOCK_SIGSTORE_TUF_REQUESTS && input.toString().startsWith(`https://tuf-repo-cdn.sigstore.dev`))
```
OR13, line 54: an allow list pattern would be better than a deny list here.

geigerzaehler: To be honest, I would like to avoid sinking more time into this, especially if a change does not make things worse than before. Or is there something I’m overlooking?

OR13: This is just a non blocking comment, especially if the pattern used here is pre-existing.

MikeMcC399:
@geigerzaehler / @aduh95 

If there any intention to continue with this PR? It seems to have stalled with no recent progress.

MikeMcC399:
@geigerzaehler 

- It's good to hear you are ready and waiting! I guess we need action from the maintainers. The last comment from @aduh95 was however in February 2025 and since then nobody from the maintainer team has been active with respect to this PR.

- I would like to see my issue https://github.com/nodejs/corepack/issues/616 able to be closed. Without a resolution, there is still exposure to Corepack failing on new pnpm releases, although we have no idea whether npm will rotate keys again, or if it won't happen again for a number of years!

MikeMcC399:
@geigerzaehler 

- Since there has been no further maintainer feedback since February 2025, and this would be essential to make any progress, I have closed the issue https://github.com/nodejs/corepack/issues/616

PR #648, opened 2025-02-11
Fix or mark false-positives in end-to-end tests

geigerzaehler:
We correct expectations against CLI output that use regular expressions. We fix tests that are easy to fix and mark tests as failing otherwise.

### DIFF
--- a/tests/main.test.ts
+++ b/tests/main.test.ts
@@ -24,7 +24,7 @@ it(`should refuse to download a package manager if the hash doesn't match`, asyn
 
     await expect(runCli(cwd, [`yarn`, `--version`])).resolves.toMatchObject({
       exitCode: 1,
-      stderr: /Mismatch hashes/,
+      stderr: expect.stringContaining(`Mismatch hashes`),
       stdout: ``,
     });
   });
@@ -35,7 +35,7 @@ it(`should refuse to download a known package manager from a URL`, async () => {
     // Package managers known by Corepack cannot be loaded from a URL.
     await expect(runCli(cwd, [`yarn@https://registry.npmjs.com/yarn/-/yarn-1.22.21.tgz`, `--version`])).resolves.toMatchObject({
       exitCode: 1,
-      stderr: /Illegal use of URL for known package manager/,
+      stderr: expect.stringContaining(`Illegal use of URL for known package manager`),
       stdout: ``,
     });
 
@@ -57,7 +57,7 @@ it.fails(`should refuse to download a known package manager from a URL in packag
 
     await expect(runCli(cwd, [`yarn`, `--version`])).resolves.toMatchObject({
       exitCode: 1,
-      stderr: /Illegal use of URL for known package manager/,
+      stderr: expect.stringContaining(`Illegal use of URL for known package manager`),
       stdout: ``,
     });
 
@@ -82,7 +82,7 @@ it(`should require a version to be specified`, async () => {
 
     await expect(runCli(cwd, [`yarn`, `--version`])).resolves.toMatchObject({
       exitCode: 1,
-      stderr: /expected a semver version/,
+      stderr: `No version specified for yarn in "packageManager" of package.json\n`,
       stdout: ``,
     });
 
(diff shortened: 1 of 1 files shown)


tests/main.test.ts
```diff
@@ -1021,76 +1021,93 @@ describe(`handle integrity checks`, () => {
...
+        stderr: expect.stringMatching(/Signature does not match/),
         stdout: ``,
       });
     });
   });
-  it(`should return an error when hash does not match without a tag`, async () => {
+
+  // FIXME: Corepack determines the latest version by asking npm instead of the
+  // configured registry. Consequently the test registry does not have the
+  // requested version.
```
aduh95, line 1051: I don't think that's true, the tests on CI do not have access to the network, it would fail if it was getting the info from the npm registry

aduh95: IIUC Corepack is trying to download the version defined in `config.json`, because we set `COREPACK_DEFAULT_TO_LATEST` to `0` on line 16. The issue is that that version doesn't exist on the custom registry

aduh95: I've opened https://github.com/nodejs/corepack/pull/649 to fix the test instead

tests/main.test.ts
```diff
@@ -1021,76 +1021,93 @@ describe(`handle integrity checks`, () => {
...
       });
+    });
+  });
+
+  // FIXME: the test registry does not implement tag fetching yet
+  it.fails(`fails on invalid signature for tag`, async () => {
+    process.env.TEST_INTEGRITY = `invalid_signature`; // See `_registryServer.mjs`
+
+    await xfs.mktempPromise(async cwd => {
       await expect(runCli(cwd, [`yarn@stable`, `--version`], true)).resolves.toMatchObject({
```
aduh95, line 1041: There seems to be something actually wrong here, even if the package manifest exposes a `stable` tag, the invalid signature is ignored nonetheless. IIUC that's due to the fact that npm signs a `.tgz` file, while the integrity for Yarn Berry is only for the `yarn.js` file.
We can test with pnpm to validate the signature verification correctly works there.

```suggestion
  it(`fails on invalid signature for tag`, async () => {
    process.env.TEST_INTEGRITY = `invalid_signature`; // See `_registryServer.mjs`

    await xfs.mktempPromise(async cwd => {
      await expect(runCli(cwd, [`pnpm@latest`, `--version`], true)).resolves.toMatchObject({
```

tests/main.test.ts
```diff
@@ -1021,76 +1021,93 @@ describe(`handle integrity checks`, () => {
...
         exitCode: 1,
-        stderr: /Signature does not match/,
+        stderr: expect.stringMatching(/Signature does not match/),
```
aduh95, line 1043: nit
```suggestion
        stderr: expect.stringContaining(`Signature does not match`),
```

tests/main.test.ts
geigerzaehler: I’ve switched it to `pnpm@latest` and it work.

tests/main.test.ts
geigerzaehler: I’ve changed it to this style everywhere.

tests/main.test.ts
```diff
@@ -1027,12 +1027,12 @@ describe(`handle integrity checks`, () => {
     await xfs.mktempPromise(async cwd => {
       await expect(runCli(cwd, [`pnpm@1.x`, `--version`], true)).resolves.toMatchObject({
         exitCode: 1,
-        stderr: /Signature does not match/,
+        stderr: expect.stringContaining(`Signature does not match`),
         stdout: ``,
       });
-      await expect(runCli(cwd, [`yarn@stable`, `--version`], true)).resolves.toMatchObject({
+      await expect(runCli(cwd, [`pnpm@latest`, `--version`], true)).resolves.toMatchObject({
```
aduh95, line 1033: FWIW I agree with the previous decision of splitting this into two tests, `1.x` is not a tag

aduh95 (approved):

PR #652, opened 2025-02-12
Use yarn nodeLinker: node-modules to make tests faster

geigerzaehler:
This change almost halves the execution time of `main.test.ts` (from 57s to 35s on my machine using NOCK_ENV=replay). I expect similar results on CI.

The end-to-end tests run corepack in a new node process about 150 times. With the pnp linker, node is asked to require `.pnp.cjs` on every startup which significantly contributes to test time.

As an alternative, I tried removing `--require=.pnp.cjs` from `NODE_OPTIONS` in `runCli`. After all, we’re running the bundled version of corepack. But this did not work because `tests/recordRequests.js` is also required which in turn needs `better-sqlite3`.

### DIFF
--- a/.yarn/patches/clipanion-npm-3.2.1-fc9187f56c.patch
+++ b/.yarn/patches/clipanion-npm-3.2.1-fc9187f56c.patch
@@ -1,10 +1,13 @@
-diff --git a/lib/platform/package.json b/lib/platform/package.json
-index 5ea9d43740d1bdb509612376c0e9ce91d20f8b20..08d4f7327dee5e605107a343d73b2254ed08ddb9 100644
---- a/lib/platform/package.json
-+++ b/lib/platform/package.json
-@@ -1,4 +1,4 @@
- {
--    "main": "./node",
-+    "main": "./node.js",
-     "browser": "./browser"
- }
+diff --git a/package.json b/package.json
+index cbf943bcc31a864f2771e457d327e5106eba9afe..84cb1ce28a4eea4dba60bb4fa6c33a08474b2d3f 100644
+--- a/package.json
++++ b/package.json
+@@ -13,7 +13,7 @@
+     "command"
+   ],
+   "version": "3.2.1",
+-  "main": "lib/advanced/index",
++  "main": "lib/advanced/index.js",
+   "license": "MIT",
+   "sideEffects": false,
+   "repository": {
--- a/.yarnrc.yml
+++ b/.yarnrc.yml
@@ -1,0 +1,2 @@
+# pnp increaeses node startup time which makes tests significantly slower.
+nodeLinker: node-modules
--- a/yarn.lock
+++ b/yarn.lock
@@ -1350,12 +1350,12 @@ __metadata:
 
 "clipanion@patch:clipanion@npm%3A3.2.1#~/.yarn/patches/clipanion-npm-3.2.1-fc9187f56c.patch":
   version: 3.2.1
-  resolution: "clipanion@patch:clipanion@npm%3A3.2.1#~/.yarn/patches/clipanion-npm-3.2.1-fc9187f56c.patch::version=3.2.1&hash=cd5507"
+  resolution: "clipanion@patch:clipanion@npm%3A3.2.1#~/.yarn/patches/clipanion-npm-3.2.1-fc9187f56c.patch::version=3.2.1&hash=687f75"
   dependencies:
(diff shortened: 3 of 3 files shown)


MikeMcC399:
- This PR would also workaround the Yarn PnP limitation described in issue https://github.com/nodejs/corepack/issues/598 and it would mean that documentation PR https://github.com/nodejs/corepack/pull/606 could be reverted.

geigerzaehler:
> I expect similar results on CI.

Yeah, for the Node 22 test on Ubuntu *total* job time drops by 40s to 1min. And on Windows it drops by 3:40 to 8min.

.yarn/patches/clipanion-npm-3.2.1-fc9187f56c.patch
```diff
@@ -1,10 +1,13 @@
...
+diff --git a/package.json b/package.json
+index cbf943bcc31a864f2771e457d327e5106eba9afe..84cb1ce28a4eea4dba60bb4fa6c33a08474b2d3f 100644
+--- a/package.json
++++ b/package.json
+@@ -13,7 +13,7 @@
+     "command"
+   ],
+   "version": "3.2.1",
+-  "main": "lib/advanced/index",
++  "main": "lib/advanced/index.js",
```
arcanis, line 10: Why those changes?

arcanis (commented):
I don't mind, it's true that we're in kind of a special case here since we're spawning a very large amount of subprocesses in our tests.

aduh95:
> since we're spawning a very large amount of subprocesses in our tests.

Could there be a way to dodge that bullet by spawning an unpatched `node` (I mean, regardless on the linker we use)? We shouldn't need any P'N'P trickery when spawning the subprocess since we're testing the bundle

.yarn/patches/clipanion-npm-3.2.1-fc9187f56c.patch
geigerzaehler: With the original patch I get the following error:

> Error: Directory import '.../corepack/node_modules/clipanion/lib/platform' is not supported resolving ES modules imported from ...corepack/node_modules/clipanion/lib/advanced/Cli.mjs

In fact, `Cli.mjs` uses `import ... from './platform'` but Node requires relative import specifiers [to always point to files](https://nodejs.org/api/esm.html#terminology). I guess this worked before because yarn’s resolver is implemented differently and supports relative package imports.

MikeMcC399:
@aduh95 

Will you merge this? It seems like a good improvement.

arcanis:
I would prefer to keep PnP, if only to have a Node.js project dogfooding the default yarn install strategy. The perf impact is unfortunate, but 30% doesn't feel too dramatic an increase that this is a must-do.

MikeMcC399:
Testing locally I observed the following timings and savings with Node.js `22.14.0` LTS running `yarn test`:

| Duration                 | Windows 11 24H2 | Ubuntu 22.04.2 LTS |
| ------------------------ | --------------- | ------------------ |
| No change (`pnp`)        | 338.35s         | 163.50s            |
| Change to `node-modules` | 179.53s         | 95.36s             |
| Savings                  | 47%             | 42%                |

Depending on needs and priorities I suggest to either merge the PR or to reject and close it.

My personal preference would be to have it merged, as it not only improves performance, it also works around issue #598.

arcanis:
I'll close it then; as a package manager author I find it somewhat reassuring to have at least one Node.js project using PnP (or loaders in general). If Corepack moves out of the Node.js org it can be revisited.

MikeMcC399:
I suggest to re-open this PR to resolve issue https://github.com/nodejs/corepack/issues/813

arcanis:
I prefer it to be fixed with https://github.com/nodejs/node/pull/62835

MikeMcC399:
> I prefer it to be fixed with [nodejs/node#62835](https://github.com/nodejs/node/pull/62835)

There are several timing dependencies concerning the viability of waiting instead of actively converting to `node-modules` right now.

This depends on:

- when Node.js releases https://github.com/nodejs/node/pull/62835 in Node.js 24. You could check with the Node.js team about when they expect to release the fix in the Active LTS version.
- when GitHub Actions updates their cached versions of Node.js 24 in `ubuntu-latest` & `macos-latest`. They may first break compatibility before fixing it.

This PR could be re-opened and merged as a short-term solution and then reverted if there is a future better way of resolving the issue.

PR #654, opened 2025-02-12
Use NOCK_ENV=replay by default

geigerzaehler:
Replaying HTTP requests makes the tests faster and aligns them with CI.

I only realized quite late that the tests get a lot faster so I thought it would be a good idea to make that the default.


geigerzaehler:
Could I get some feedback on this?

aduh95:
I'm not sure, it would feel a bit "too magical" for the default behavior to not match reality. Is it so bad to have to set the variable manually / go through the network?

PR #655, opened 2025-02-13
Fix eslint warning and reject warnings on CI

geigerzaehler:


aduh95:
I've opened https://github.com/nodejs/corepack/pull/657 to address it a bit "better", I wish TS was better at handling this kind of cases.

sources/main.ts
```diff
@@ -1,5 +1,4 @@
 import {BaseContext, Builtins, Cli}    from 'clipanion';
-import type {UsageError}               from 'clipanion';
```
aduh95, line 2:
```suggestion
import type {UsageError}               from 'clipanion';

```

Issue #597, opened 2025-01-14
ci vitest fails "handle integrity checks" on Windows Node.js 23


MikeMcC399:
## Issue

When running `corepack yarn test` on Windows with Node.js 23.x the test fails with:

>  FAIL  tests/main.test.ts > handle integrity checks > should return an error when signature does not match with a tag

The issue does not occur on Windows with Node.js 18, 20 or 22, for example Node.js 22.13.0 LTS.

This is a problem preventing CI updating to testing on Node.js 23. At this time it is not identified as an end-user issue.

## Steps to reproduce

Windows 11 24H2
Node.js `v23.6.0`
Use Drive `C:` for clone

It is also reproducible in GitHub runners `windows-2022` and `windows-2025`

```shell
git clone https://github.com/nodejs/corepack
cd corepack
corepack yarn add better-sqlite3@11.7.2 -D # until PR #594 is merged
corepack yarn build
corepack yarn run vitest --run
```

or to only run the test section which fails:

```shell
corepack yarn run vitest -t "handle integrity checks" --run
```

## Logs

```text
$ corepack yarn add better-sqlite3@11.7.2 -D # until PR #594 is merged
➤ YN0000: · Yarn 4.3.1
➤ YN0000: ┌ Resolution step
➤ YN0000: └ Completed in 0s 277ms
➤ YN0000: ┌ Post-resolution validation
➤ YN0086: │ Some peer dependencies are incorrectly met by dependencies; run yarn explain peer-requirements for details.
➤ YN0000: └ Completed
➤ YN0000: ┌ Fetch step
➤ YN0013: │ 424 packages were added to the project (+ 132.08 MiB).
➤ YN0000: └ Completed in 21s 597ms
➤ YN0000: ┌ Link step
➤ YN0000: │ ESM support for PnP uses the experimental loader API and is therefore experimental
➤ YN0007: │ better-sqlite3@npm:11.7.2 must be built because it never has been before or the last one failed
➤ YN0007: │ esbuild@npm:0.21.5 must be built because it never has been before or the last one failed
➤ YN0000: └ Completed in 13s 9ms
➤ YN0000: · Done with warnings in 34s 979ms
```

```text
$ corepack yarn build

  dist\lib\corepack.cjs  925.6kb

⚡ Done in 619ms
All shims have been generated.
```

```text
$ corepack yarn run vitest --run

 RUN  v2.0.3 C:/Users/mikem/github/corepack

 ✓ tests/config.test.ts (1)
[...]

PR #660, opened 2025-02-14
ci: enable node 23 ci tests for windows

MikeMcC399:
## Issue

- Due to issue https://github.com/nodejs/corepack/issues/597 Windows was excluded from testing in Node.js 23 inside [.github/workflows/ci.yml](https://github.com/nodejs/corepack/blob/main/.github/workflows/ci.yml).

The above issue is no longer reproducible.

## Change

Remove the exclusion of running CI tests on Windows under Node.js 23 from the workflow [.github/workflows/ci.yml](https://github.com/nodejs/corepack/blob/main/.github/workflows/ci.yml).

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -55,13 +55,6 @@ jobs:
           - macos-latest
           - windows-latest
 
-      # Temporarily skipping Node.js 23 under Windows due to issue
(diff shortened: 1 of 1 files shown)


aduh95:
I'm guessing it's related to https://github.com/nodejs/corepack/pull/645

Issue #659, opened 2025-02-14
Corepack not using specified package manager version (in this example, pnpm)


jpbnetley:
I previously had an issue where it could not install the new version of pnpm, as stated here: https://github.com/nodejs/corepack/issues/612
I updated corepack with `npm i -g corepack@latest`, for in case that is relevant.

I have a specified version of pnpm in the package json.
But when I run `pnpm -v` it shows an outdated version.

The repo I am referencing is here: https://github.com/jpbnetley/joke-of-the-day
In the package json, I have
```
"packageManager": "pnpm@10.4.0+sha512.6b849d0787d97f8f4e1f03a9b8ff8f038e79e153d6f11ae539ae7c435ff9e796df6a862c991502695c7f9e8fac8aeafc1ac5a8dab47e36148d183832d886dd52",
```

But when I check the version in the terminal for pnpm, it is `v9.6.0`

Here is some additional info (I am using WSL for my dev env)

```sh
joke-of-the-day git:(dev) corepack -v
0.31.0
➜  joke-of-the-day git:(dev) node -v
v23.7.0
➜  joke-of-the-day git:(dev) pnpm -v
9.6.0
➜  joke-of-the-day git:(dev) corepack enable
➜  joke-of-the-day git:(dev) pnpm -v        
9.6.0
➜  joke-of-the-day git:(dev) corepack up
Installing pnpm@10.4.0 in the project...

Lockfile is up to date, resolution step is skipped
Packages: +5
+++++
Progress: resolved 5, reused 5, downloaded 0, added 5, done

devDependencies:
+ @testing-library/react 16.2.0
+ @typescript-eslint/eslint-plugin 8.24.0
+ eslint-plugin-import 2.31.0
+ eslint-plugin-jest 28.11.0

> joke-of-the-day@0.1.0 prepare /home/ubi/dev_projects/joke-of-the-day
> husky || exit 0


Done in 1.3s
➜  joke-of-the-day git:(dev) pnpm -v
9.6.0
```

Any advice is appreciated 😄

aduh95:
Can you try setting `DEBUG=corepack` in your env?

jpbnetley:
> Can you try setting `DEBUG=corepack` in your env and report the output?

I added the suggested env, please let me know if I did it incorrectly.

```sh
joke-of-the-day git:(dev) DEBUG=corepack
➜  joke-of-the-day git:(dev) echo $DEBUG   
corepack
[...]

PR #661, opened 2025-02-14
docs: caveats of multiple package managers

jpbnetley:
Added documentation regarding multiple installations of a package manager with different versions.

Originally the issue was opened [here](https://github.com/nodejs/corepack/issues/659)

And it was suggested to maybe add documentation to outline the potential pitfall [here](https://github.com/nodejs/corepack/issues/659#issuecomment-2658993332).

If there is a better way to word the documentation I am contributing, please advice.
I am open to suggestions. 😄 

I am also not sure under what heading I should add it under, for now I opted to add it under `Troubleshooting` heading.

issue src #659 

I just realized now that it was suggested to add the docs to pnpm, not  corepack.
If that is still preferred, this pr can be closed 😓

### DIFF
--- a/README.md
+++ b/README.md
@@ -333,6 +333,14 @@ There are a wide variety of networking issues that can occur while running
   `curl [URL]` (ipv4) and `curl -6 [URL]` (ipv6) from your shell.
 - Check your proxy settings (see [Environment Variables](#environment-variables)).
 
+### Package manager
+If corepack is enabled, corepack will ensure that the correct package manager is used when specified in the package.lock.json  
+However, a potential pitfall is that if there are multiple installations of the same package manager.  
+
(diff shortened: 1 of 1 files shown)


MikeMcC399:
@jpbnetley 

The intention of my suggestion was to have a warning added to the https://pnpm.io/installation documentation because it lists multiple ways of installing pnpm, including using `curl` and Corepack.  The pnpm documentation website would be the first place to put a warning about using different installation methods at the same time, which could cause version mismatch and functional confusion.

This is a special problem of pnpm, with its variety of installation methods, so it belongs more in the pnpm documentation.

jpbnetley:
Thanks @MikeMcC399, Noted 😄

PR #662, opened 2025-02-17
Remove temporary home directories in test teardown

geigerzaehler:
Remove temporary home directories in test teardown. I actually ran out of space a couple of times and before I could remove the temporary directories I had to change the permissions.

### DIFF
--- a/tests/main.test.ts
+++ b/tests/main.test.ts
@@ -11,16 +11,16 @@ import {runCli}                                      from './_runCli';
 
 
 beforeEach(async () => {
+  const home = await xfs.mktempPromise();
+
   // `process.env` is reset after each tests in setupTests.js.
-  process.env.COREPACK_HOME = npath.fromPortablePath(await xfs.mktempPromise());
+  process.env.COREPACK_HOME = npath.fromPortablePath(home);
   process.env.COREPACK_DEFAULT_TO_LATEST = `0`;
-});
 
-afterEach(async () => {
-  const home = npath.toPortablePath(process.env.COREPACK_HOME!);
-  await xfs.removePromise(home, {recursive: true});
+  return async () => {
+    await xfs.removePromise(home, {recursive: true});
+  };
 });
-
 
 it(`should refuse to download a package manager if the hash doesn't match`, async () => {
   await xfs.mktempPromise(async cwd => {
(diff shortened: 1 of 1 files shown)


tests/main.test.ts
```diff
@@ -480,6 +486,16 @@ it(`should support disabling the network accesses from the environment`, async (
 });
 
 describe(`read-only and offline environment`, () => {
+  let home: PortablePath;
+  beforeEach(() => {
+    home = npath.toPortablePath(process.env.COREPACK_HOME!);
+  });
+
+  afterEach(async () => {
```
aduh95, line 494:
```suggestion
  afterEach(async () => {
    const home = npath.toPortablePath(process.env.COREPACK_HOME!);

```

tests/main.test.ts
```diff
@@ -480,6 +486,16 @@ it(`should support disabling the network accesses from the environment`, async (
...
+    await xfs.chmodPromise(ppath.join(home, `lastKnownGood.json`), 0o644);
+    await xfs.chmodPromise(home, 0o755);
```
aduh95, line 496: Could we add it only on the test that needs it? You can use `t.after` I think

tests/main.test.ts
```diff
@@ -16,6 +16,12 @@ beforeEach(async () => {
   process.env.COREPACK_DEFAULT_TO_LATEST = `0`;
 });
 
+afterEach(async () => {
+  const home = npath.toPortablePath(process.env.COREPACK_HOME!);
+  await xfs.removePromise(home, {recursive: true});
+});
+
```
aduh95, line 23: Let's return a teardown function instead, should be cleaner. Could you also add this to the other tests please?
```suggestion
  return () => xfs.removePromise(home, {recursive: true});
});
```

tests/main.test.ts
geigerzaehler: Oh yeah, this doesn’t make sense. I think I wanted to reuse home initially. Fixed in https://github.com/nodejs/corepack/pull/662/commits/da69ce3d4c579290a8e262350ed955858e586183

tests/main.test.ts
geigerzaehler: `onTestFinished` does not work because the cleanups from all `beforeEach` block are run before the hooks registered with `onTestFinished`. So the cleanup tries to remove `home` before the permissions are reverted and removal throws an exception.

The cleanup code is scoped to a `describe` block and required by all tests in that block so it doesn’t run unneeded.

tests/main.test.ts
geigerzaehler: Yes, that’s definitely cleaner. Fixed in https://github.com/nodejs/corepack/pull/662/commits/da69ce3d4c579290a8e262350ed955858e586183

aduh95:
Let's apply that to all the test files, not just `main.test.ts`

Issue #465, opened 2024-04-22
feat: store `packageManager` property also inside lock file


ruscon:
Our current flow:
We use `docker` + `pnpm fetch`
```
COPY --chown=node:node pnpm-*.yaml .npmrc ./
RUN corepack enable
RUN --mount=type=cache,id=pnpm,target=/pnpm/store pnpm fetch --loglevel=warn --ignore-scripts
```

This means we don't cache the `package.json` file before the `pnpm fetch` execution.
This allows us to worry about `package.json` changes.
But corepack will always install the latest version of `pnpm` instead of a specific one.

If you store the `packageManager` property in a lock file, this will solve the problem.

styfle:
This sounds like it should be implemented in `pnpm` and not `corepack`

ruscon:
@styfle Perhaps I don’t know how corepack works under the hood, but most likely some kind of layer is needed for interaction between corepack and the package manager, so that the package manager itself updates its lock file?

aduh95:
Have you considered setting a `packageManager` in your `package.json`? That way, Corepack would pick up the correct version when run inside the container. I'm not sure I understand why `package.json` couldn't be the "lock file" in this case.

ruscon:
@aduh95 The question is related to how the `docker` cache and `pnpm fetch` work.
`pnpm fetch` and `npm ci` can be run without a `package.json` file, which helps cache the installation of npm packages inside the docker image even if you make changes to the `package.json` file.

aduh95:
I see, pardon my ignorance, I should have read the docs. From what I can see, pnpm docs recommend downloading a specific version of pnpm with cURL globally. The equivalent with corepack would be to run `corepack install -g pnpm@8.x`. So IIUC, what you're requesting would be to have a way to skip the command and have `pnpm fetch` select the correct version. I don't know enough pnpm internals to tell how easy/hard that would be to implement, but I agree that would require changes on pnpm side – so that `package.json` it generates contains the correct `"packageManager"` field.

PR #668, opened 2025-02-28
feat: let `.corepack.env` be a lock file

### LINKED ISSUE #95: Support for semver ranges?

### LINKED ISSUE #402: `packageManager` field is too limited

### LINKED ISSUE #682: Feature request: read `packageManager` and `devEngines` fields from outside of package.json
Use case:

When developing some projects which reject to add `packageManager` and `devEngines` fields to its package.json, it would be not easy to use correct pacakge manager and package manager version, even though I've enabled corepack.

if corepack reads `packageManager` and `devEngines` fields from outside of package.json like .corepack.env, it would be convenient.

aduh95:
With https://github.com/nodejs/corepack/pull/642 and https://github.com/nodejs/corepack/pull/643 landed, we can consider using `.corepack.env` as a lockfile. If the `package.json` defines a `devEngines.packageManager`, we can accept an env variable that defines the exact version Corepack should be using; if that version is put in a `.corepack.env` (Node.js 20+ users only), it's effectively a lockfile.

I'm not a fan of the env variable name chosen, happy to use a different one.

Fixes: https://github.com/nodejs/corepack/issues/402
Fixes: https://github.com/nodejs/corepack/issues/95
Fixes: https://github.com/nodejs/corepack/issues/682

### DIFF
--- a/README.md
+++ b/README.md
@@ -286,6 +286,10 @@ same major line. Should you need to upgrade to a new major, use an explicit
   package manager, and to not update the Last Known Good version when it
   downloads a new version of the same major line.
 
+- `COREPACK_DEV_ENGINES_${UPPER_CASE_PACKAGE_MANAGER_NAME}` can be set to give
+  Corepack a specific version matching the range defined in `package.json`'s
+  `devEngines.packageManager` field.
+
 - `COREPACK_ENABLE_AUTO_PIN` can be set to `0` to prevent Corepack from
   updating the `packageManager` field when it detects that the local package
   doesn't list it. In general we recommend to always list a `packageManager`
--- a/sources/specUtils.ts
+++ b/sources/specUtils.ts
@@ -103,6 +103,17 @@ function parsePackageJSON(packageJSONContent: CorepackPackageJSON) {
 
     debugUtils.log(`devEngines.packageManager defines that ${name}@${version} is the local package manager`);
 
+    const localEnvKey = `COREPACK_DEV_ENGINES_${packageManager.name.toUpperCase()}`;
+    const localEnvVersion = process.env[localEnvKey];
+    if (localEnvVersion) {
+      debugUtils.log(`Environment defines that ${name}@${localEnvVersion} is the local package manager`);
+
+      if (!semverSatisfies(localEnvVersion, version))
+        warnOrThrow(`"${localEnvKey}" environment variable is set to ${JSON.stringify(localEnvVersion)} which does not match the value defined in "devEngines.packageManager" for ${JSON.stringify(name)} of ${JSON.stringify(version)}`, onFail);
+
+      return `${name}@${localEnvVersion}`;
+    }
+
     if (pm) {
       if (!pm.startsWith?.(`${name}@`))
         warnOrThrow(`"packageManager" field is set to ${JSON.stringify(pm)} which does not match the "devEngines.packageManager" field set to ${JSON.stringify(name)}`, onFail);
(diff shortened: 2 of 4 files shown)


styfle:
To be clear, this is only going to update `.corepack.env` if it already exists.

Its not going to create `.corepack.env` if it does not exist, right?

aduh95:
> To be clear, this is only going to update `.corepack.env` if it already exists.
> 
> Its not going to create `.corepack.env` if it does not exist, right?

Yes correct, if it exists and contains the env key – so it’s currently opt-in, I think we could discuss whether we want to flip that to opt-out once the `parseEnv` thing is no longer experimental.

arcanis (commented):
Why do we need that considering that `packageManager` can already be used as a lock for `devEngines.packageManager`?

aduh95:
> Why do we need that considering that `packageManager` can already be used as a lock for `devEngines.packageManager`?

So one can git ignore it I believe is the ask

arcanis:
I'm not sure I follow the use case. Why would they use Corepack if they don't want to lock the version in the project?

aduh95:
@zanminkian could you give an example of use case where putting a .gitignored `.corepack.env` with `COREPACK_ENABLE_AUTO_PIN=0` would not fulfill your needs?

styfle:
> Then disable auto pinning? I don't think this behavior should have been made the default anyway.

I tend to agree. I believe the default auto-pinning was added as a response to earlier TSC meetings where this was a recommendation to making corepack stable and default enabled with node distributions.

Now that corepack is no longer going to be bundled with node, we could consider disabling the auto pinning behavior.

khokm:
Linked issues: https://github.com/nodejs/corepack/issues/465
I support moving the packageManager field out of package.json into a separate file. This can be especially useful for Docker workflows like:

```dockerfile
RUN corepack enable

# Copy package.json so corepack can read the packageManager field
COPY package.json ./package.json
RUN corepack prepare

COPY pnpm-lock.yaml pnpm-workspace.yaml ./
# pnpm fetch is a time-consuming operation
RUN --mount=type=cache,id=pnpm,target=/pnpm/store pnpm fetch
```

Currently, any change in package.json that does not affect the build—for example, adding a new entry in scripts—forces Docker to invalidate the cache.

An alternative would be using corepack install -g pnpm@<version>, but that requires keeping the packageManager field in package.json and the pnpm version in the Dockerfile in sync.

Having a separate file (like a proposed .corepack.env) containing only packageManager version would theoretically change much less frequently than package.json, reducing unnecessary Docker cache invalidation.

My current workaround is instead of copying package.json, to read only `packageManager` from it and pass it as an ARG

escalonn:
docker wants corepack to be able to discover packageManager outside package.json, because otherwise, you have to either
1. copy package.json before anything else, so if you are incrementing your version there on every deploy, you can't do any layer caching, or
2. have nice caching, but you have to let corepack pick a random version of your package manager for the lock-file-based install, then get the correct version for subsequent commands after the package.json is copied in

khokm:
> docker wants corepack to be able to discover packageManager outside package.json, because otherwise, you have to either
> 
> 1. copy package.json before anything else, so if you are incrementing your version there on every deploy, you can't do any layer caching, or
> 2. have nice caching, but you have to let corepack pick a random version of your package manager for the lock-file-based install, then get the correct version for subsequent commands after the package.json is copied in

This can probably be solved by using multiple stages:

```dockerfile
FROM node:20-alpine AS corepack-extract

WORKDIR /app

# jq is required to filter JSON content
RUN apk add --no-cache jq

# Copy full package.json only in this stage
COPY package.json .

# Create a minimal package.json containing only packageManager
# Key point: Corepack reads only this file, so it is isolated
# from all other changes in package.json
RUN jq '{packageManager}' package.json > package.corepack.json

FROM node:20-alpine AS deps

WORKDIR /app

# Use minimal package.json so Corepack sees only packageManager
# and does not react to unrelated changes in dependencies or metadata
COPY --from=corepack-extract /app/package.corepack.json ./package.json

# Enable Corepack (disabled by default in some Node images)
RUN corepack enable

PR #670, opened 2025-03-06
Removes npm from the possible options

arcanis:
I never intended for Corepack to offer npm as an option _without its maintainers enabling it themselves_. It was supported through an explicit `corepack enable npm` mostly to make sure that whatever design we picked would keep npm in mind, so that in the event they'd one day be interested to join us it'd just be a flip of a switch. It wasn't a requirement (npm users already have their nice developer experience, my goal was to improve Yarn's and pnpm's devx at no cost for anyone).

To clarify that and make it _super duper clear_ that making Corepack stable won't impact npm users in any shape or form, I propose this PR to remove `npm` from the possible options. It's no longer possible to activate it using `corepack enable npm`. I erased any mention of it, including from the tests.

Also, addressing my own comment:
https://github.com/nodejs/corepack/pull/418#issuecomment-1982978140

> get formal approval from someone from the actual npm team. As far as I'm aware you're working on a competing product, which makes you kind of the wrong person to make this request (part of why I didn't merge my own https://github.com/nodejs/corepack/pull/407).

This was requested/suggested by TSC members unrelated to Yarn / pnpm. Since this is a Node.js project, I don't have objection to them requesting this change.

Closes #418

### DIFF
--- a/README.md
+++ b/README.md
@@ -4,8 +4,7 @@
 
 Corepack is a zero-runtime-dependency Node.js script that acts as a bridge
 between Node.js projects and the package managers they are intended to be used
-with during development. In practical terms, **Corepack lets you use Yarn, npm,
-and pnpm without having to install them**.
+with during development. In practical terms, **Corepack lets you use Yarn and pnpm without having to install them**.
 
 ## How to Install
 
@@ -71,7 +70,7 @@ See [`CONTRIBUTING.md`](./CONTRIBUTING.md).
 ### When Building Packages
 
 Just use your package managers as you usually would. Run `yarn install` in Yarn
-projects, `pnpm install` in pnpm projects, and `npm` in npm projects. Corepack
+projects and `pnpm install` in pnpm projects. Corepack
 will catch these calls, and depending on the situation:
 
 - **If the local project is configured for the package manager you're using**,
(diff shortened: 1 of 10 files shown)


BridgeAR (approved):
LGTM

This seems to supersede and close https://github.com/nodejs/corepack/pull/418

arcanis:
> I disagree. I dont think npm has any way to manage versions per working directory like you can do with corepack.

Then they are the best people to lobby. Corepack wasn't intended to improve their devx. Don't get me wrong, I personally think it would be a good thing for npm to leverage Corepack, but it's not my castle and I don't think Yarn and pnpm users should suffer so much uncertainty because of that.

Npm (and the Node TSC) will always have the ability to discuss eventual future plans through follow-ups, but they won't need me in that discussion 🙂

aduh95 (changes requested):
My previous concerns remain, I don’t think we gain anything from it, and unless the crurent npm maintainers are asking for this, I don’t think we should do it

styfle:
>  I don't think Yarn and pnpm users should suffer so much delay and uncertainty because of that.

Can you elaborate? Why would those users suffer because of npm being available through corepack? 🤔

arcanis:
> Can you elaborate? Why would those users suffer because of npm being available through corepack? 🤔

I don't think I said that? What I did say/mean is that I share the opinion that the mere presence of npm in the Corepack debate is unfortunately making it radioactive. I believe it is in part why this proposal hasn't been going anywhere despite strong repeated support from the community.

I hope that by opening this PR it'll be clear that marking Corepack stable will **not** change anything to npm users - and perhaps quell some concerns.

styfle:
> the mere presence of npm in the Corepack debate is unfortunately making it radioactive.

Can you provide evidence of this? Who is debating and what are they saying specifically?

PR #709, opened 2025-05-10
Disables auto-pinning by default

### LINKED ISSUE #485: `COREPACK_ENABLE_AUTO_PIN` should default to `0`

arcanis:
I believe Corepack shouldn't have any negative impact on projects that aren't configured to use it. It should only be a net positive. Changing a file from the project has negative impacts as it leads to confusion ("why is this file changing?"), frustration ("I need to revert the file after every command!"), and sometimes breakages (pristine checks).

I didn't object much to auto-pinning at the time it was introduced because it was a request made by people from the TSC, and I was interested in compromise seeking. Considering the TSC declined to push Corepack further, I see less value in that and I'd like to revisit this design.

Fixes #485

Closes #552

### DIFF
--- a/README.md
+++ b/README.md
@@ -286,8 +286,8 @@ same major line. Should you need to upgrade to a new major, use an explicit
   package manager, and to not update the Last Known Good version when it
   downloads a new version of the same major line.
 
-- `COREPACK_ENABLE_AUTO_PIN` can be set to `0` to prevent Corepack from
-  updating the `packageManager` field when it detects that the local package
+- `COREPACK_ENABLE_AUTO_PIN` can be set to `1` to instruct Corepack to
+  update the `packageManager` field when it detects that the local package
   doesn't list it. In general we recommend to always list a `packageManager`
   field (which you can easily set through `corepack use [name]@[version]`), as
   it ensures that your project installs are always deterministic.


aduh95:
Duplicate of https://github.com/nodejs/corepack/pull/552?

arcanis:
> Duplicate of #552?

I didn't see it before but yeah, same intent. It supersedes it though; it also updates the tests.

MikeMcC399:
I would appreciate this PR moving forward!

Just yesterday I had to ask somebody to revert a commit because they had accidently added a Yarn Classic packageManager to a repo that uses npm as package manager.

Removing the default auto-pinning would help prevent these sorts of mistakes.

trivikr (commented):
LGTM. Waiting to see it in newer version of corepack, taking it more close to 1.0.0

crazyair:
How to subscribe to the distribution? I want to upgrade as soon as possible

MikeMcC399:
@arcanis / @aduh95 

This PR doesn't contain a `feat:` commit message, so it hasn't been picked up by `release-please` and PR https://github.com/nodejs/corepack/pull/699.

Could you fix it up so it can be released please? I also submitted #713 to request this.

PR #714, opened 2025-05-26
feat: Adds guard to avoid stepping on Yarn's feet

arcanis:
I'm working on an experimental alternative to Corepack that also replaces the binary *(while I'm pretty sure this experimental alternative will be released in some way, I don't plan to remove Yarn support from Corepack; this is more intended to avoid having all our eggs in the same basket)*.

To make sure it interacts correctly with Corepack, this PR adds a check to avoid accidentally erasing the experimental Yarn jumper when running `corepack enable` / `corepack disable` (with the rational that an "official" jumper would always necessarily know how to fetch Yarn versions, whereas Corepack *could* conceivably lag in implementing support for future url patterns; so it should have priority).

### DIFF
--- a/tests/Enable.test.ts
+++ b/tests/Enable.test.ts
@@ -1,7 +1,7 @@
 import {Filename, ppath, xfs, npath}                                    from '@yarnpkg/fslib';
 import {delimiter}                                                      from 'node:path';
 import process                                                          from 'node:process';
-import {describe, beforeEach, it, expect}                               from 'vitest';
+import {describe, beforeEach, it, expect, test}                         from 'vitest';
 
 import {Engine}                                                         from '../sources/Engine';
 import {SupportedPackageManagers, SupportedPackageManagerSetWithoutNpm} from '../sources/types';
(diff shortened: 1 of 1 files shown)


tests/Enable.test.ts
```diff
@@ -87,4 +91,42 @@ describe(`EnableCommand`, () => {
       await expect(sortedEntries).resolves.toEqual(expectedEntries.sort());
     });
   });
+
+  testNotWindows(`should overwrite existing files`, async () => {
```
aduh95, line 95: https://vitest.dev/api/#test-skipif
```suggestion
  test.skipIf(process.platform === `win32`)(`should overwrite existing files`, async () => {
```

sources/commands/Disable.ts
```diff
@@ -70,6 +70,13 @@ export class DisableCommand extends Command<Context> {
   async removePosixLink(installDirectory: string, binName: string) {
     const file = path.join(installDirectory, binName);
     try {
+      const currentTarget = await fs.promises.realpath(file);
+
+      if (binName.includes(`yarn`) && currentTarget.match(/[/\\]switch[/\\]bin[/\\]/)) {
```
aduh95, line 75: We should run the `realpath` only for Yarn
```suggestion
      if (binName.includes(`yarn`) && /[/\\]switch[/\\]bin[/\\]/.test(await fs.promises.realpath(file))) {
```

PR #739, opened 2025-09-17
feat: Allow token to be sent on download

### LINKED ISSUE #733: Still 401 when retrieving a package manager binary from artifactory as an authenticated NPM registry
This refers to closed #275

Sorry for ME being the bearer of bad news, but this is still no longer working.

My setup:
artifactory as private registry, requests have to be authorized.
nodejs 22.18.0
corepack 0.34.0

Authentication is setup using env variables:
```
COREPACK_NPM_REGISTRY=https://some.host/artifactory/api/npm/registry
COREPACK_NPM_TOKEN=<token>
```

When I do:
`corepack install pnpm@latest -g`

There happen to be 3 requests, the third is unauthorized:
```
https://some.host/artifactory/api/npm/registry/pnpm
https://some.host/artifactory/api/npm/registry/pnpm/10.14.0
https://some.host/artifactory/api/npm/registry/
```

One reason might be https://github.com/nodejs/corepack/blob/0b492c9c97e4b3d5e9a139eeedb931860127b470/sources/httpUtils.ts#L32

which is incorrect in my view, because `input.origin` happens to be `https://some.host` without a path, so this expression can never be true.

But even then - the 3rd request is crap anyway, it should be something like that, but it isn't:
`https://some.host/artifactory/api/npm/registry/pnpm/-/pnpm-10.14.0.tgz`

Why?

This seems to be related to #725

BobLuursema:
When using Artifactory the registry is set to for example: `https://my-company.example.org/artifactory/api/npm/my-npm`. Authentication is mandatory in our setup. But the token will only be sent if the hostname matches with the value in the environment variable. This blocks the download since the variable also includes the path portion.

I am not sure how precise you want to be with this check. If needed I can also change it to parse out the hostname from the environment variable for a more secure check.

Fixes #733 

In addition, I didn't see a test that checks this check. I can add it if you'd like.

### DIFF
--- a/sources/httpUtils.ts
+++ b/sources/httpUtils.ts
@@ -29,7 +29,7 @@ async function fetch(input: string | URL, init?: RequestInit) {
     input.username = input.password = ``;
   }
 
-  if (input.origin === (process.env.COREPACK_NPM_REGISTRY || DEFAULT_NPM_REGISTRY_URL) && process.env.COREPACK_NPM_TOKEN) {
+  if ((process.env.COREPACK_NPM_REGISTRY || DEFAULT_NPM_REGISTRY_URL).includes(input.origin) && process.env.COREPACK_NPM_TOKEN) {
     headers =  {
       ...headers,
       authorization: `Bearer ${process.env.COREPACK_NPM_TOKEN}`,
--- a/sources/httpUtils.ts
+++ b/sources/httpUtils.ts
@@ -20,8 +20,8 @@ async function fetch(input: string | URL, init?: RequestInit) {
   const username: string | undefined = input.username || process.env.COREPACK_NPM_USERNAME;
   const password: string | undefined = input.password || process.env.COREPACK_NPM_PASSWORD;
 
-  if (username && password) {
-    const encodedCreds = Buffer.from(`${username}:${password}`, `utf8`).toString(`base64`);
+  if (username || password) {
+    const encodedCreds = Buffer.from(`${username ?? ''}:${password ?? ''}`, `utf8`).toString(`base64`);
     headers = {
       ...headers,
       authorization: `Basic ${encodedCreds}`,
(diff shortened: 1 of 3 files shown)


sources/httpUtils.ts
```diff
@@ -29,7 +29,7 @@ async function fetch(input: string | URL, init?: RequestInit) {
     input.username = input.password = ``;
   }
 
-  if (input.origin === (process.env.COREPACK_NPM_REGISTRY || DEFAULT_NPM_REGISTRY_URL) && process.env.COREPACK_NPM_TOKEN) {
+  if ((process.env.COREPACK_NPM_REGISTRY || DEFAULT_NPM_REGISTRY_URL).includes(input.origin) && process.env.COREPACK_NPM_TOKEN) {
```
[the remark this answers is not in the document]
BobLuursema: No, `DEFAULT_NPM_REGISTRY_URL` is set as a `const` in `npmRegistryUtils.ts` so that should be impossible.

aduh95 (approved):

sources/httpUtils.ts
```diff
@@ -20,17 +20,18 @@ async function fetch(input: string | URL, init?: RequestInit) {
   const username: string | undefined = input.username || process.env.COREPACK_NPM_USERNAME;
   const password: string | undefined = input.password || process.env.COREPACK_NPM_PASSWORD;
 
-  if (username || password) {
-    headers =  {
+  if (username && password) {
```
styfle, line 23: Now that both username and password are required to set the authorization, should we warn if one is set and not the other?

BobLuursema: Oh, I am dumb, this `if` would not pass with an empty string of course 🤦‍♂️

aduh95:
CI is failing, can you take a look?

victor-perez:
I also created a fix for this problem in #743, which is a smaller change that only addresses the bug. This PR, on the other hand, also updates the way authentication is handled and adds some useful tests.

Maybe we can first fix the bug and then follow up with the overall improvements from this PR, since I see it still needs some test fixes.

It would be great if we could fix this bug quickly. After the Shai-Hulud attack, I expect more companies will move to registries that let them scan, block, and have more control over the packages being downloaded. These registries can also include paths.

PR #743, opened 2025-09-30
Fix incorrect registry origin check

victor-perez:
Update the registry origin check to use the correct URL object for comparison.

### DIFF
--- a/sources/httpUtils.ts
+++ b/sources/httpUtils.ts
@@ -29,9 +29,9 @@ async function fetch(input: string | URL, init?: RequestInit) {
     input.username = input.password = ``;
   }
 
-  const registry = new URL(process.env.COREPACK_NPM_REGISTRY || DEFAULT_NPM_REGISTRY_URL);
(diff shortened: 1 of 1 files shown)


sources/httpUtils.ts
```diff
@@ -29,7 +29,9 @@ async function fetch(input: string | URL, init?: RequestInit) {
     input.username = input.password = ``;
   }
 
-  if (input.origin === (process.env.COREPACK_NPM_REGISTRY || DEFAULT_NPM_REGISTRY_URL) && process.env.COREPACK_NPM_TOKEN) {
+  const registry = new URL(process.env.COREPACK_NPM_REGISTRY || DEFAULT_NPM_REGISTRY_URL);
+
+  if (input.origin === registry.origin && process.env.COREPACK_NPM_TOKEN) {
```
aduh95, line 34: nit: we could avoid calling the `URL` constructor if `COREPACK_NPM_TOKEN` is not set / falsy
```suggestion
  const registry = process.env.COREPACK_NPM_TOKEN && new URL(process.env.COREPACK_NPM_REGISTRY || DEFAULT_NPM_REGISTRY_URL);

  if (registry && input.origin === registry.origin) {
```

victor-perez:
Runner didn't started correctly for 1 test @aduh95

PR #764, opened 2025-11-07
Fixes concurrent downloads of the same Yarn binary

arcanis:
We had a check allowing the `rename` operation to fail for tarballs, but forgot to do the same for the single JS file downloaded in the Yarn codepath. As a result the operation could fail when two processes are downloading the exact same version. I got this error on Netlify when trying to install multiple git dependencies from the same repository.

Unfortunately no tests, as it's a race condition so not sure I can easily test that without making the tests flaky 🫤

### DIFF
--- a/sources/nodeUtils.ts
+++ b/sources/nodeUtils.ts
@@ -5,7 +5,7 @@ export interface NodeError extends Error {
 }
 
 export function isNodeError(err: any): err is NodeError {
-  return `code` in err;
+  return !!err?.code;
 }
 
 export function isExistError(err: NodeError) {


sources/nodeUtils.ts
```diff
@@ -4,6 +4,10 @@ export interface NodeError extends Error {
   code: string;
 }
 
+export function isExistError(err: NodeError) {
```
aduh95: We can use this as an opportunity to reduce the number of `as` in the codebase
```suggestion
export function isNodeError(err: unknown): err isNodeError {
  return !!err?.code;
}

export function isExistError(err: NodeError) {
```

sources/nodeUtils.ts
```diff
@@ -4,6 +4,14 @@ export interface NodeError extends Error {
   code: string;
 }
 
+export function isNodeError(err: any): err is NodeError {
+  return `code` in err;
```
aduh95: That's unsafe, we'll get `Cannot use 'in' operator to search for 'code' in undefined` if `err` is nullish
```suggestion
  return !!err?.code;
```

PR #767, opened 2025-11-10
fix: Revert to readlink in generatePosixLink to avoid unnecessary unlinks

polyomino24:
## Context / Background

While running `corepack enable` without `sudo` in a CircleCI docker environment(cimg/node:22.20.0), the build failed with an EACCES: permission denied, unlink '/usr/local/bin/pnpm' error.

This error was caused by the recent change #761  from readlink to realpath in generatePosixLink.

https://github.com/nodejs/corepack/pull/761/files#diff-c91b034b85c9610409bf66de5d28560a06f1285ee1962f238fac2c3652a15e88L86-R88

## Problem
Using fs.promises.realpath in generatePosixLink returns the fully resolved, absolute path of the symlink target.

However, the symlink variable used in the comparison is a relative path calculated by path.relative. This results in a comparison between an absolute path and a relative path, causing the currentSymlink !== symlink check to always evaluate to true.

https://github.com/nodejs/corepack/blob/cf0ddc18cd949b984ea70a02e9fb57831d2b0b3a/sources/commands/Enable.ts#L83-L98

### DIFF
--- a/sources/Engine.ts
+++ b/sources/Engine.ts
@@ -244,7 +244,7 @@ export class Engine {
    * project using the default package managers, and configure it so that we
    * don't need to ask again in the future.
    */
-  async findProjectSpec(initialCwd: string, locator: Locator | LazyLocator, {transparent = false}: {transparent?: boolean} = {}): Promise<Descriptor> {
+  async findProjectSpec(initialCwd: string, locator: Locator | LazyLocator, {transparent = false, binaryVersion}: {transparent?: boolean, binaryVersion?: string | null} = {}): Promise<Descriptor> {
     // A locator is a valid descriptor (but not the other way around)
     const fallbackDescriptor = {name: locator.name, range: `${locator.reference}`};
 
@@ -293,7 +293,7 @@ export class Engine {
         }
 
         case `Found`: {
-          const spec = result.getSpec();
+          const spec = result.getSpec({enforceExactVersion: !binaryVersion});
           if (spec.name !== locator.name) {
             if (transparent) {
               if (typeof locator.reference === `function`)
@@ -344,7 +344,7 @@ export class Engine {
       };
     }
 
-    const descriptor = await this.findProjectSpec(cwd, fallbackLocator, {transparent: isTransparentCommand});
+    const descriptor = await this.findProjectSpec(cwd, fallbackLocator, {transparent: isTransparentCommand, binaryVersion});
 
     if (binaryVersion)
       descriptor.range = binaryVersion;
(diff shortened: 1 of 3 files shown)


sources/commands/Enable.ts
```diff
@@ -83,9 +83,9 @@ export class EnableCommand extends Command<Context> {
     const symlink = path.relative(installDirectory, path.join(distFolder, `${binName}.js`));
 
     if (fs.existsSync(file)) {
-      const currentSymlink = await fs.promises.realpath(file);
+      const currentSymlink = await fs.promises.readlink(file);
```
aduh95: The issue with that is that it's going to throw if `file` is not a symlink. We should probably add tests otherwise this kind of regression is bound to happen again in the future

polyomino24: I've pushed a fix that uses lstat to check isSymbolicLink() before calling readlink, and it now correctly unlinks non-symlink files too.

I also added the tests.

sources/commands/Enable.ts
```diff
@@ -83,17 +83,23 @@ export class EnableCommand extends Command<Context> {
...
+          console.warn(`${binName} is already installed in ${file} and points to a Yarn Switch install - skipping`);
+          return;
+        }
+
+        if (currentSymlink !== symlink) {
+          await fs.promises.unlink(file);
+        } else {
+          return;
+        }
       } else {
```
aduh95: We can simplify this I believe
```suggestion
        if (currentSymlink === symlink) return;
      }
```

sources/commands/Enable.ts
```diff
@@ -83,17 +83,23 @@ export class EnableCommand extends Command<Context> {
     const symlink = path.relative(installDirectory, path.join(distFolder, `${binName}.js`));
 
     if (fs.existsSync(file)) {
-      const currentSymlink = await fs.promises.readlink(file);
+      const stats = await fs.promises.lstat(file);
```
aduh95: `existsSync` and `lstat` duplicates the work, can we simplify this:

```js
const stats = fs.lstatSync(file, { throwIfNoEntry: false });
if (stats) {
```

tests/Enable.test.ts
```diff
@@ -125,4 +126,50 @@ describe(`EnableCommand`, () => {
       expect(file).toBe(`hello`);
     });
   });
+
+  test.skipIf(process.platform === `win32`)(`should not re-link if binaries are already correct`, async () => {
+    await xfs.mktempPromise(async cwd => {
+      await makeBin(cwd, `corepack` as Filename);
+
+      process.env.PATH = `${npath.fromPortablePath(cwd)}${delimiter}${process.env.PATH}`;
+
```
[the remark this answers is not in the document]
polyomino24: You are right. removed

PR #800, opened 2026-03-03
Prefer stable versions when resolving range (#774)

### LINKED ISSUE #774: Corepack incorrectly selects prerelease/dev versions of pnpm when no version is specified
> After some discussion on the pnpm issue (https://github.com/pnpm/pnpm/issues/10214), it turns out the problem is not caused by a faulty pnpm release, pnpm v11 _**has changed its binary entry path from bin/pnpm.cjs to bin/pnpm.mjs.**_, so the development build 11.0.0-dev.1005 is not faulty. Corepack is expecting the old path and fails.
> 
> So the issue appears to be:
> 
> - Corepack’s version resolution prefers a dev build over the latest stable
> - Corepack seems to rely on a hardcoded pnpm binary path instead of reading the correct bin path from package.json.
> 
> As a workaround I'm currently using `corepack use pnpm@10.22.0 `or `corepack use pnpm@latest `(or latest-9, latest-10 etc.)
> 
> Leaving this update here so the root cause is clearer. 

 _Originally posted by @tkesici in [#772](https://github.com/nodejs/corepack/issues/772#issuecomment-3569230701)_

### Problem
When running `corepack use pnpm`, Corepack unexpectedly selects and installs a prerelease development build of pnpm (11.0.0-dev.1005) instead of the latest stable version .

This leads to failure because pnpm v11 is not stable yet and its structure differs from the current pnpm 10.x series.

### Expected Behavior
If no version is specified, Corepack should prefer the latest stable release of pnpm (e.g 10.x)

### Actual Behavior

`corepack use pnpm` resolves to: Installing pnpm@11.0.0-dev.1005...

### Environment

**Node.js: 20 (node:20-alpine)**

**Corepack: 0.34.4**

**OS: Alpine Linux**

AMDphreak:
This PR fixes #774 by ensuring that Corepack prefers the latest stable version over newer prereleases when resolving a range (like * or latest), unless a prerelease is explicitly requested.\n\n### Problem\nWhen multiple versions match a range, Corepack currently just sorts them with semver.rcompare. Since prereleases (e.g., 11.0.0-dev...) are considered higher than the latest stable (e.g., 10.4.1), Corepack incorrectly selects the prerelease. This causes issues when prereleases have breaking changes in binary paths or behavior.\n\n### Fix\nIn 
esolveDescriptor, we now split matching versions into stable and prerelease sets. We prefer the highest stable version if one exists, and only fallback to prereleases if the requested range specifically targets them (checked via isPrereleaseRange helper).

### DIFF
--- a/sources/Engine.ts
+++ b/sources/Engine.ts
@@ -1,22 +1,23 @@
-import {UsageError}                                           from 'clipanion';
-import fs                                                     from 'fs';
-import path                                                   from 'path';
-import process                                                from 'process';
-import semverRcompare                                         from 'semver/functions/rcompare';
-import semverValid                                            from 'semver/functions/valid';
-import semverValidRange                                       from 'semver/ranges/valid';
+import { UsageError } from 'clipanion';
+import fs from 'fs';
+import path from 'path';
+import process from 'process';
+import semverRcompare from 'semver/functions/rcompare';
+import semverValid from 'semver/functions/valid';
+import semverParse from 'semver/functions/parse';
+import semverValidRange from 'semver/ranges/valid';
 
-import defaultConfig                                          from '../config.json';
+import defaultConfig from '../config.json';
 
-import * as corepackUtils                                     from './corepackUtils';
-import * as debugUtils                                        from './debugUtils';
-import * as folderUtils                                       from './folderUtils';
-import type {NodeError}                                       from './nodeUtils';
-import * as semverUtils                                       from './semverUtils';
-import * as specUtils                                         from './specUtils';
-import {Config, Descriptor, LazyLocator, Locator}             from './types';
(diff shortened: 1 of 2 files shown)


MikeMcC399 (changes requested):
There are a lot of unrelated formatting changes in this PR, which means that it fails linting, and is difficult to review.

The following should fix the formatting:

```shell
corepack yarn lint --fix
```

Please reformat the text in the original post which includes `\n` characters instead of actual newlines.

This repo uses `release-please` so it would need a commit message like `fix:`

An additional test case would help check that the PR fulfils its intended purpose.

I'm not a maintainer, so I can't approve or process this PR further. It would however be good for issue #774 to be fixed, as it has caused me problems in the past.

Issue #799, opened 2026-03-03
corepack use fails when preinstall scripts invoke the package manager


AMDphreak:
This issue describes a deadlock/race condition in the lifecycle management of Corepack on Windows...\n\n### Scenario\n1. Project has a package.json with a preinstall script that invokes the package manager (e.g., pnpm dlx only-allow pnpm).\n2. User runs corepack use pnpm@latest.\n3. Corepack updates the manifest and initiates an internal install.\n4. The preinstall script is triggered. Because Corepack is currently re-linking or locking the global shims (e.g. pnpm.exe) to the new version, the child process spawned to run the script cannot find pnpm in the PATH or hits a file lock.\n5. The installation crashes, leaving the global shims in a broken state.\n\n### Related Information\n- This is exacerbated by #774, as selecting a prerelease with different binary paths immediately breaks the shim for subsequent lookups.\n\n### Proposed Fix\nI have submitted a PR (#798) that addresses this by:\n1. Preferring stable versions in 
esolveDescriptor if no prerelease is explicitly requested.\n2. Setting COREPACK_PACKAGE_MANAGER_LOCATOR in 
unVersion. Shims and sub-invocations can use this environment variable to bipass version lookups and use the currently active version directly, avoiding the PATH/shim race condition.

MikeMcC399:
Do please format this issue so that it is human-readable!

PR #801, opened 2026-03-03
Fix preinstall deadlock on Windows by using COREPACK_PACKAGE_MANAGER_LOCATOR (#799)

AMDphreak:
This PR fixes a deadlock/race condition on Windows where project preinstall scripts that invoke the package manager (e.g., pnpm dlx only-allow pnpm) can crash during corepack use because the global shim is being re-linked or locked by Corepack. This is documented in #799.\n\n### Fix\nIn 
unVersion, we now set the COREPACK_PACKAGE_MANAGER_LOCATOR environment variable. In xecutePackageManagerRequest, we check this variable. If it matches the requested binary, we use it directly to bypass the project specification lookup and shim resolution.\n\n### Why this helps\nThis ensures that child processes sparked by the package manager during lifestyle scripts use the already-resolved version immediately, avoiding any interaction with external shims that might be in a transient or locked state.

### DIFF
--- a/sources/Engine.ts
+++ b/sources/Engine.ts
@@ -1,22 +1,22 @@
-import {UsageError}                                           from 'clipanion';
-import fs                                                     from 'fs';
-import path                                                   from 'path';
-import process                                                from 'process';
-import semverRcompare                                         from 'semver/functions/rcompare';
-import semverValid                                            from 'semver/functions/valid';
-import semverValidRange                                       from 'semver/ranges/valid';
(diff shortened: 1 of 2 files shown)


MikeMcC399 (changes requested):
As far as I can judge, this PR fails linting and tests. Both would need to be fixed.
It also introduces a new environment variable `COREPACK_PACKAGE_MANAGER_LOCATOR` which is undocumented in the README.

Again, note that I am not a maintainer.

PR #811, opened 2026-04-18
fix: stop manifest search on devEngines package managers

### LINKED ISSUE #779: Asymmetry between searching for closest package.json when packageManager is defined vs when devEngines.packageManager is defined after #643
TL;DR I believe that in #643 [one part of one line of code](https://github.com/nodejs/corepack/blame/v0.34.5/sources/specUtils.ts#L173) should have been changed from `!selection.data.packageManager` to `!parsePackageJSON(selection.data)` in order to make a consistent change. 

As this was not done, there is now asymmetry in lookup of "package.json that defines package manager to be used", package.json that uses `packageManager` is always respected, package.jsons that use `devEngine.packageManager` are only respected when they are top-most package.json.

I'd like a check from @aduh95 as the author of #643, I am happy to provide a PR that cleans up the current asymmetry. 

----

In the current code (encountered / verified experimentally on v0.34.5 with `DEBUG=corepack` ), the algorithm defined in https://github.com/nodejs/corepack/blame/v0.34.5/sources/specUtils.ts#L166-L238, way before #643 (and #642 which was merged immediately after), loops through package.json files, starting in current directory, and stops looping & updating the `selection` when package.json with `packageManager` field is found. 

Otherwise, the selection will end up the last package.json (closest to filesystem root) encountered -- and in that case, we know `packageManager` is not set in it either, so when the code tries again! check on it on line 237, we return 'NoSpec' and will fallback to global PM version. This bit is important.

In #643, the basic support for reading `devEngines.packageManager` was added, including checking it's semver against `packageManager` field if both are present. It was tied into the original logic by replacing the code that was reading `selection.data.packageManager` with a call of new function that takes both fields into account: https://github.com/nodejs/corepack/commit/b4562688513f23e37e37b0d69a0daff33ca84c8d#diff-388b281fcee55f16c4f20b24ef733cc04cc2605d2e9753cffbb0ceb9ce8e6ef9L120-R193 (scroll to `sources/specUtils.ts` lines L120 / R 193).

However, the other critical piece, which says "look for package.json that defines package manager version to be used" , implemented in terms of "keep looping until package.json with `packageManager` is encountered) remained unchanged: (https://github.com/nodejs/corepack/blame/v0.34.5/sources/specUtils.ts#L173.

This is causing an asymmetry during the package.json looping/scanning. If there is a package.json with `packageManager: <nonnullvalue>`, the lookup will stop there. However, if the package.json contains (a valid) `devEngines.packageManager` field, the lookup will not stop and continue looping. 

If there is no package.json above, the new feature for devEngines works "by accident", because the `selection` is not cleaned on each iteration, so https://github.com/nodejs/corepack/blame/v0.34.5/sources/specUtils.ts#L236 loads information from it. 

However, in all other cases, the `devEngines.packageManager` value will be completely ignored.

e.g. in my case, this was inside a monorepo that had another package.json above it, closer to the root (it was npm monorepo package this package was was NOT part of). This other package.json was the last seen, it was stuck in `selection` and when fed into the rest of the function in https://github.com/nodejs/corepack/blame/v0.34.5/sources/specUtils.ts#L236 it decides the outcome. Since it did not have any PM definition, 'NoSpec' was returned it fallback to global PM.

other potential outcome is that if there would be any package.json above, with `packageManager` set, the lookup would stop on that file, as said, ignoring `devEngines.packageManager` in the first file.

Note: the root cause / logic from https://github.com/nodejs/corepack/issues/560 does not seem to apply here.

P.S.: I'd go for packageManager field, but there is high change of accidentally using non-corepack-aliased-npm on this project, and npm does only care about devEngines.packageManger, so hence my choice of using this approach.

----

My conclusion after this little deep dive: the current behavior seems hardly desirable as it is inconsistent, and should be unified in a way that the lookup stops for packages defining either `packageManager` of `devEngines.packageManager`. There is likely missing functional test coverage for this use cases of feature #643 that could be improved. Moreover, the pre-existing implementation choice where outdated `selection` data is kept is bug-prone and should be refactored.

michkot:
fixes #779 

**Summary**

This branch is intentionally a small stack, not a single behavioral change.

The main fix is the manifest-selection bug from #779: when scanning parent directories, `loadSpec` now stops at the nearest `package.json` that defines either `packageManager` or `devEngines.packageManager`. Before that, the scan only stopped for `packageManager`, so a closer manifest using only `devEngines.packageManager` could be skipped and a parent manifest could win instead.

The branch then adds two explicit follow-up behavior changes that became clearer while fixing the lookup:

- top-level `packageManager: ""` and `packageManager: null` are now treated as defined but invalid values of the nearest manifest, so lookup stops there and reports an invalid `"packageManager"` instead of treating those values as absent and falling through to a parent `package.json`
- `Engine.findProjectSpec` now honors non-error `onFail` for package-manager name mismatches when the selected project package manager comes only from `devEngines.packageManager`; in those cases, `onFail: warn` / `ignore` can fall back to the requested package manager instead of throwing

The branch also improves invalid `"packageManager"` errors so they mention which field in the selected `package.json` was used.

**Decision table**

The table below describes the outcome for the nearest selected manifest. It intentionally ignores transparent-command behavior, which can still choose fallback for separate reasons.

| packageManager | devEngines.packageManager | Effective onFail | Selected-manifest result                     | Engine.findProjectSpec result                        |
| -------------- | ------------------------- | ---------------- | -------------------------------------------- | ---------------------------------------------------- |
| absent         | absent                    | n/a              | continue scanning upward                     | if nothing else is found: NoSpec -> use fallback     |
| valid          | absent                    | n/a              | stop and use packageManager                  | enforce that package manager                         |
| invalid        | absent                    | n/a              | stop and report invalid packageManager       | throw invalid packageManager error                   |
| absent         | valid                     | error            | stop and use devEngines.packageManager       | enforce that package manager; mismatch throws        |
| absent         | valid                     | warn / ignore    | stop and use devEngines.packageManager       | mismatch falls back to requested package manager     |
| absent         | invalid                   | error            | stop and report invalid devEngines           | throw invalid devEngines.packageManager error        |
| absent         | invalid                   | warn / ignore    | stop at nearest manifest and return NoSpec   | use fallback                                         |
| valid          | valid and matching        | any              | stop and use packageManager                  | enforce packageManager                               |
| valid          | valid but mismatching     | error            | stop and report field mismatch               | throw mismatch error                                 |
| valid          | valid but mismatching     | warn / ignore    | stop and use packageManager                  | enforce packageManager                               |
| valid          | invalid                   | error            | stop and report invalid devEngines           | throw invalid devEngines.packageManager error        |
| valid          | invalid                   | warn / ignore    | stop and use packageManager                  | enforce packageManager                               |

**Tests**

- keep one CLI regression for the mixed-field case where a child `devEngines.packageManager` must win over a parent `packageManager`
- add `Engine.findProjectSpec` coverage for:
  - nearest child `devEngines.packageManager` when the parent has no package-manager field
  - nearest child `devEngines.packageManager` over a parent `packageManager`
  - nearest invalid child `devEngines.packageManager` with `onFail: warn`, which must stop the search and use fallback rather than a parent `packageManager`
  - nearest child `packageManager: ""`
  - nearest child `packageManager: null`

**Notes**

- This branch changes behavior for `packageManager: ""` and `packageManager: null`: they no longer behave like an absent field during upward lookup.
- Follow-up suggestion: `onFail` should probably not affect structurally invalid `devEngines.packageManager` definitions, for example when `name` is not a string or `version` is not a valid semver range. Those look like malformed configuration rather than mismatch-policy choices, so they likely should always throw. The mismatch cases can still remain controlled by `onFail`.

**Commit ordering**

- the first two commits are pure repros, so reviewers can inspect the failing scenarios before reading the fix
- the next functional commits are ordered from root cause to follow-up semantics: first nearest-manifest selection, then explicit `packageManager: "" | null` handling, then the `onFail: warn` fallback behavior, then the error-path improvement
- the CLI coverage reduction is kept as its own test-shaping commit so the single retained end-to-end repro stays obvious
- the final docs and refactor commits are intentionally last, so reviewers can read behavioral changes without rename / comment / control-flow cleanup noise mixed into them

### DIFF
--- a/sources/Engine.ts
+++ b/sources/Engine.ts
@@ -302,7 +302,7 @@ export class Engine {
               debugUtils.log(`Falling back to ${fallbackDescriptor.name}@${fallbackDescriptor.range} in a ${spec.name}@${spec.range} project`);
               return fallbackDescriptor;
             } else {
-              throw new UsageError(`This project is configured to use ${spec.name} because ${result.target} has a "packageManager" field`);
+              throw new UsageError(`This project is configured to use ${spec.name} because ${result.target} has a "${result.sourceField}" field`);
             }
           } else {
             debugUtils.log(`Using ${spec.name}@${spec.range} as defined in project manifest ${result.target}`);
--- a/sources/Engine.ts
+++ b/sources/Engine.ts
@@ -229,19 +229,20 @@ export class Engine {
   }
 
   /**
-   * Locates the active project's package manager specification.
+   * Locates the package manager specification from nearest package.json having one.
    *
    * If the specification exists but doesn't match the active package manager,
    * an error is thrown to prevent users from using the wrong package manager,
    * which would lead to inconsistent project layouts.
    *
-   * If the project doesn't include a specification file, we just assume that
-   * whatever the user uses is exactly what they want to use. Since the version
-   * isn't specified, we fallback on known good versions.
+   * If none of the package.json files have specification,
+   * or a package.json had specification which was invalid but allowed failure,
+   * we just assume that whatever the user uses is exactly what they want to use.
+   * Since the version isn't specified, we fallback on known good versions.
    *
-   * Finally, if the project doesn't exist at all, we ask the user whether they
-   * want to create one in the current project. If they do, we initialize a new
-   * project using the default package managers, and configure it so that we
+   * Finally, if the package.json doesn't exist at all, we ask the user whether they
+   * want to create one in the current working directory. If they do, we initialize a new
+   * package.json using the default package managers, and configure it so that we
    * don't need to ask again in the future.
    */
   async findProjectSpec(initialCwd: string, locator: Locator | LazyLocator, {transparent = false, binaryVersion}: {transparent?: boolean, binaryVersion?: string | null} = {}): Promise<Descriptor> {
--- a/sources/Engine.ts
+++ b/sources/Engine.ts
@@ -258,66 +258,66 @@ export class Engine {
     if (process.env.COREPACK_ENABLE_STRICT === `0`)
       transparent = true;
 
-    while (true) {
-      const result = await specUtils.loadSpec(initialCwd);
+    const result = await specUtils.loadSpec(initialCwd);
 
-      switch (result.type) {
-        case `NoProject`: {
-          if (typeof locator.reference === `function`)
-            fallbackDescriptor.range = await locator.reference();
+    switch (result.type) {
+      case `NoProject`: {
+        if (typeof locator.reference === `function`)
+          fallbackDescriptor.range = await locator.reference();
 
-          debugUtils.log(`Falling back to ${fallbackDescriptor.name}@${fallbackDescriptor.range} as no project manifest were found`);
-          return fallbackDescriptor;
+        debugUtils.log(`Falling back to ${fallbackDescriptor.name}@${fallbackDescriptor.range} as no project manifest were found`);
+        return fallbackDescriptor;
+      }
+
+      case `NoSpec`: {
+        if (typeof locator.reference === `function`)
+          fallbackDescriptor.range = await locator.reference();
+
+        if (process.env.COREPACK_ENABLE_AUTO_PIN === `1`) {
+          const resolved = await this.resolveDescriptor(fallbackDescriptor, {allowTags: true});
+          if (resolved === null)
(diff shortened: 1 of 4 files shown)


michkot:
@aduh95 May I ask you for a code review?

michkot:
@aduh95 anything I can do to increase chance this is reviewed/merged in any form?

Issue #813, opened 2026-04-18
CI tests fail under Node.js 24.15.0 LTS (EBADF)


MikeMcC399:
## Current behavior

Running `typecheck`, `lint` or `test`, as used in [.github/workflows/ci.yml](https://github.com/nodejs/corepack/blob/main/.github/workflows/ci.yml), fails when run locally under Node.js 24.15.0 LTS with:

> Error: EBADF: bad file descriptor, fstat

This issue ~~can also be expected to appear soon~~ appears under GitHub Actions and causes CI workflow failure ~~if / when~~ since the cached Node.js versions in GitHub Actions runner images are updated to include Node.js 24.15.0 LTS.

## Expected behavior

It should be possible to test Corepack successfully under Node.js Active LTS, including under the latest release Node.js 24.15.0.

## Versions

Corepack `0.34.7`
Ubuntu `24.04.4` LTS
Node.js `24.15.0` (Active LTS)
Yarn `4.11.0`

## Steps to reproduce

Install Node.js 24.15.0 LTS, then execute:

```shell
git clone https://github.com/nodejs/corepack
cd corepack
npm uninstall yarn -g # if previously installed
npm install corepack -g
corepack yarn install
corepack yarn build
corepack yarn typecheck
corepack yarn lint
corepack yarn test
```

## Logs

```text
$ corepack yarn install
corepack yarn build
corepack yarn test
➤ YN0000: · Yarn 4.11.0
➤ YN0000: ┌ Resolution step
➤ YN0000: └ Completed
➤ YN0000: ┌ Fetch step
➤ YN0000: └ Completed in 0s 250ms
➤ YN0000: ┌ Link step
➤ YN0000: │ ESM support for PnP uses the experimental loader API and is therefore experimental
➤ YN0007: │ better-sqlite3@npm:11.10.0 must be built because it never has been before or the last one failed
➤ YN0007: │ esbuild@npm:0.27.7 must be built because it never has been before or the last one failed
➤ YN0000: └ Completed in 1m 1s
➤ YN0000: · Done with warnings in 1m 2s

  dist/lib/corepack.cjs  954.6kb

⚡ Done in 66ms
All shims have been generated.
node:fs:391
  const stats = binding.fstat(fd, false, undefined, true /* shouldNotThrow */);
                        ^

Error: EBADF: bad file descriptor, fstat
    at tryStatSync (node:fs:391:25)
    at readFileSync (node:fs:447:17)
[...]

PR #816, opened 2026-04-22
chore(tests): add support for missing localhost in DNS

aduh95:
If for whatever reason, DNS cannot retrieve localhost, fallback to the default unspecified address (i.e. `::` or `0.0.0.0`).

### DIFF
--- a/tests/_registryServer.mjs
+++ b/tests/_registryServer.mjs
@@ -187,7 +187,15 @@ if (process.env.AUTH_TYPE === `PROXY`) {
     });
   });
   proxy.listen(0, `localhost`);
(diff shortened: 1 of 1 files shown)


MikeMcC399:
@aduh95 

Your PR has hit the CI-failure issue described in https://github.com/nodejs/corepack/issues/813. Since the runner version is random, you may be able to get CI to succeed by re-running, and hoping that the job is allocated an old runner version.

I've just submitted https://github.com/nodejs/corepack/pull/817 as a mitigation.

aduh95:
Giving up on that, it's probably fine to expect `localhost` to be resolvable

PR #863, opened 2026-06-19
chore(*): refactor to use built-in `--strip-types` Node.js flag

aduh95:
With support for Node.js 20 being dropped, we no longer need `tsx` and can rely on the built-in Node.js type stripping

### DIFF
--- a/eslint.config.mjs
+++ b/eslint.config.mjs
@@ -17,6 +17,7 @@ export default [
         name: `fetch`,
         message: `Use fetch from sources/httpUtils.ts`,
       }],
+      '@typescript-eslint/consistent-type-imports': `error`,
       '@typescript-eslint/no-unused-vars': [`error`, {
         caughtErrors: `none`,
       }],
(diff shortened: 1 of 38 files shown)


MikeMcC399:
@aduh95 

https://github.com/nodejs/corepack/actions/runs/27810046674/job/82298074570

`corepack yarn build` shows:

```console
(node:2425) [MODULE_TYPELESS_PACKAGE_JSON] Warning: Module type of file:///home/runner/work/corepack/corepack/mkshims.ts is not specified and it doesn't parse as CommonJS.
Reparsing as ES module because module syntax was detected. This incurs a performance overhead.
To eliminate this warning, add "type": "module" to /home/runner/work/corepack/corepack/package.json.
(Use `node --trace-warnings ...` to show where the warning was created)
All shims have been generated.
```

That should probably be addressed.

aduh95:
I'm fine with the performance overhead, as long as that doesn't affect the distributed version

MikeMcC399:
> I'm fine with the performance overhead, as long as that doesn't affect the distributed version

The PR removes the dependency on `tsx` and it is green in CI, so I don't have any other points to raise.

PR #880, opened 2026-07-27
fix: Error when upgrading pnpm to latest in devEngines.packageManager

### LINKED ISSUE #874: Error when upgrading pnpm to latest in `devEngines.packageManager`
If pnpm is previously set at a specific version such as `11.12.0` in `devEngines.packageManager`:

`package.json`

```json
{
  "devEngines": {
    "packageManager": {
      "name": "pnpm",
      "version": "11.12.0+sha512.820a6fbd0d9f04c226638002aead1e45340a9139dd5dc077c1d83ef44aa2481c8eb6637b4c9aa696a3c7e35ba818e49cf27213e5f2b91138d09b7a3e26e898ba"
    }
  }
}
```

Then upgrading using `corepack use pnpm@11.13.0` (or `corepack use pnpm@latest`) fails with a confusing error:

```bash
$ corepack use pnpm@11.13.0
Installing pnpm@11.13.0 in the project...
Usage Error: The requested version of pnpm@11.13.0+sha512.88d94724d8f2e6c186744a5584c6e59ecac869ec7ba15e9cb4cd628e8dc7066820b2481d8ee3b51ea8da323a7378068aa58c556a3720d32b7c20a051d088363a does not match the devEngines specification (pnpm@11.12.0+sha512.820a6fbd0d9f04c226638002aead1e45340a9139dd5dc077c1d83ef44aa2481c8eb6637b4c9aa696a3c7e35ba818e49cf27213e5f2b91138d09b7a3e26e898ba)

$ corepack use <pattern>
```

It appears to be caused by Corepack attempting to add a `packageManager` field for `11.13.0` which conflicts with `devEngines.packageManager`'s version of `11.12.0`.

Some projects only want to rely on `devEngines.packageManager`, instead of also using `packageManager` or duplicating the fields.

Could Corepack instead update the version in `devEngines.packageManager` and not attempt to add `packageManager`?

## Related issues

It appeared at first that this was related to issue #729, but I think my issue here is narrower - about a specific failure case.

thiagobarbosa:
Closes #874

## Problem

When a project pins an exact version in `devEngines.packageManager` and has no packageManager
field, running corepack use <name>@<newVersion> fails with a confusing error, because      
Corepack compares the new version against the pinned devEngines range instead of just      
updating it.

## Changes
When `devEngines.packageManager` is the only source of truth (no `packageManager` field, same
package manager name, and an exact version), update `devEngines.packageManager.version` in
place instead of validating against it and adding a separate `packageManager` field.

## Tests
Added a test in `tests/Use.test.ts` reproducing the issue: it pins `devEngines.packageManager.version` to `2.1.0`, runs `use yarn@2.4.3`, and asserts the `devEngines` version is bumped while no `packageManager` field is added.


aduh95:
I'd say it's expected, if `devEngines` sets a specific version, Corepack should report trying to use a different one, no? `corepack use` is special in the fact that it does write the `package.json`, so I can see how updating the `devEngines` is tempting, but arguably Corepack does not own this field.
I guess it depends how niche is the usecase of setting a specific version in `devEngines`, I would expect that ranges are more common – in which case, Corepack already does what the use expect IMO.

Issue #884, opened 2026-08-05
Support signature verification against a custom COREPACK_NPM_REGISTRY


opswithranjan:
### Summary

When `COREPACK_NPM_REGISTRY` points at a private registry that **re-signs** packages with its own key, `corepack prepare` / `corepack install` fail with:

```
Usage Error: The package was not signed by any trusted keys
```

Corepack fetches package metadata and tarballs from the configured registry, but it only ever verifies signatures against the **npm public keys bundled in `config.json`** (or a manually supplied `COREPACK_INTEGRITY_KEYS`). It never consults the configured registry's own published keys, so a package signed by that registry can never be trusted.

### Where it happens

`sources/npmRegistryUtils.ts`, `verifySignature`:

```js
const {npm: trustedKeys} = process.env.COREPACK_INTEGRITY_KEYS
  ? JSON.parse(process.env.COREPACK_INTEGRITY_KEYS)
  : defaultConfig.keys;   // <-- always npmjs keys, even when COREPACK_NPM_REGISTRY is a private registry
```

The registry URL and auth are already known here (`COREPACK_NPM_REGISTRY`, `COREPACK_NPM_TOKEN`, etc.), but the registry's `/-/npm/v1/keys` endpoint is never fetched.

### Why this can't reasonably be worked around by users

- There's no way for many consumers (e.g. Dependabot jobs) to inject `COREPACK_INTEGRITY_KEYS` into the environment.
- `.corepack.env` is not loaded by the `prepare` / `install` subcommands (#741), so committing config doesn't help.
- The result is broken installs for anyone behind a re-signing private registry (Cloudsmith, and similar), across every corepack-driven tool (npm/pnpm/yarn activation, and downstream tools like Dependabot and mise).

### Proposed change

When a non-default `COREPACK_NPM_REGISTRY` is configured and `COREPACK_INTEGRITY_KEYS` is **not** explicitly set, corepack should fetch that registry's signing keys from `<registry>/-/npm/v1/keys` (the standard npm registry keys endpoint) and trust them **in addition to** the bundled npm keys.

Trusted-key resolution would become:

| Config | Trusted keys |
|---|---|
| `COREPACK_INTEGRITY_KEYS` set | that value (unchanged, explicit override wins) |
| default npmjs registry | bundled npm keys (unchanged) |
| custom `COREPACK_NPM_REGISTRY` | registry's `/-/npm/v1/keys` **+** bundled npm keys |

[...]

Issue #741, opened 2025-09-18
corepack use does not load COREPACK_NPM_REGISTRY from .corepack.env


Gadzy:
PROBLEM
========

`corepack use` does not load `COREPACK_NPM_REGISTRY` from `.corepack.env`

CONFIGURATION
=============

- macOS v15.6.1
- NodeJS v22.19.0
- corepack v0.34.0

DESCRIPTION
===========

`https://registry.npmjs.org` is blocked by my company.
We use a private repository, and want to enforce it's use in our `Node.js` projects via a `.corepack.env` file.

```.corepack.env
COREPACK_NPM_REGISTRY=https://our.private.repo
```

```
$ corepack use pnpm
Internal Error: Error when performing the request to https://registry.npmjs.org/pnpm; for troubleshooting help, see https://github.com/nodejs/corepack#troubleshooting
    at fetch (/Users/gadzy/.local/share/fnm/node-versions/v22.19.0/installation/lib/node_modules/corepack/dist/lib/corepack.cjs:22051:11)
    at process.processTicksAndRejections (node:internal/process/task_queues:105:5)
    at async fetchAsJson (/Users/gadzy/.local/share/fnm/node-versions/v22.19.0/installation/lib/node_modules/corepack/dist/lib/corepack.cjs:22065:20)
    at async fetchAvailableVersions (/Users/gadzy/.local/share/fnm/node-versions/v22.19.0/installation/lib/node_modules/corepack/dist/lib/corepack.cjs:22009:20)
    at async fetchAvailableVersions2 (/Users/gadzy/.local/share/fnm/node-versions/v22.19.0/installation/lib/node_modules/corepack/dist/lib/corepack.cjs:22146:14)
    at async /Users/gadzy/.local/share/fnm/node-versions/v22.19.0/installation/lib/node_modules/corepack/dist/lib/corepack.cjs:23001:25
    at async Promise.all (index 1)
    at async Engine.resolveDescriptor (/Users/gadzy/.local/share/fnm/node-versions/v22.19.0/installation/lib/node_modules/corepack/dist/lib/corepack.cjs:22998:22)
    at async UseCommand.execute (/Users/gadzy/.local/share/fnm/node-versions/v22.19.0/installation/lib/node_modules/corepack/dist/lib/corepack.cjs:23469:22)
[...]

PR #885, opened 2026-08-05
feat: trust signing keys published by a custom npm registry

opswithranjan:
Refs #884

## Problem

When `COREPACK_NPM_REGISTRY` points at a registry that re-signs the packages it serves, Corepack rejects the package outright:

```
Usage Error: The package was not signed by any trusted keys: {
  "signatures":  [ { "keyid": "SHA256:5mJAtUoc…" } ],
  "trustedKeys": [ "SHA256:jl3bwswu…", "SHA256:DhQ8wR5…" ]
}
```

Corepack fetches metadata and tarballs from the configured registry, but `verifySignature` only ever compares against the npm keys bundled in `config.json` (or a manually supplied `COREPACK_INTEGRITY_KEYS`). The registry's own signing keys are never consulted, so a package it signed can never verify.

This isn't practically workaroundable for affected users: many consumers (CI images, Dependabot jobs) can't inject `COREPACK_INTEGRITY_KEYS` into the environment, and `.corepack.env` isn't loaded by the `prepare` / `install` subcommands (#741).

## What this changes

When the bundled keys don't already cover the signature **and** a non-default `COREPACK_NPM_REGISTRY` is configured **and** `COREPACK_INTEGRITY_KEYS` isn't set, Corepack now fetches that registry's keys from `<registry>/-/npm/v1/keys` (the endpoint defined by [npm's registry-signature spec](https://docs.npmjs.com/about-registry-signatures)) and trusts them in addition to the bundled ones.

| Configuration | Trusted keys | Extra request |
| --- | --- | --- |
| `COREPACK_INTEGRITY_KEYS` set | that value (override still wins) | no |
| default npm registry | bundled npm keys | no |
| custom registry, signature matches bundled keys | bundled npm keys | **no** |
| custom registry, signature doesn't match | bundled **+** registry-published keys | yes, once per registry |

**Added to, not replaced.** A proxying registry serves a mix: versions it has stored are signed with its own key, while versions it hasn't pulled yet are passed through from upstream with npm's original signature intact. Corepack can be asked for either, so both key sets have to be trusted. Merging also means the change can only ever add trust, where replacing the bundled keys could stop legitimately npm-signed packages from verifying.

**Lazy on purpose.** The request only happens on the path that currently hard-fails, so every configuration that works today issues exactly the same requests as before.

**Fails safe, never open.** If the endpoint is unreachable, returns no `keys` array, returns an empty one, or every entry is malformed, Corepack keeps verifying against the bundled keys and reports the usual error. It never falls back to skipping verification.

## Trust model

The keys are fetched from the same registry, over the same channel and with the same credentials, that serves the tarball Corepack is about to unpack and execute. Consulting its published keys is strictly less privilege than the code execution Corepack already grants it, so this doesn't widen the trust boundary. It also matches what the npm CLI does: it verifies against the configured registry's keys, not a hardcoded set.

## Testing

`tests/npmRegistryUtils.test.ts` grows from 7 to 19 cases, covering: verifying against a registry-published key; sending registry credentials on the keys request; trailing-slash normalisation; ignoring malformed entries; fetching at most once per registry; **not** fetching when no custom registry is set / `COREPACK_INTEGRITY_KEYS` is set / the bundled keys already match; and falling back rather than skipping verification when the endpoint fails or publishes nothing usable.

Verified end to end against Cloudsmith that re-signs what it serves. The registry below is public, so this reproduces with no credentials:

```bash
export COREPACK_NPM_REGISTRY="https://npm.cloudsmith.io/opswithranjan/npm"
unset COREPACK_INTEGRITY_KEYS
COREPACK_HOME=$(mktemp -d) corepack prepare npm@11.16.0 --activate
```

| Command (identical env, only the branch differs) | `main` | this branch |
| --- | --- | --- |
| `corepack prepare npm@11.16.0 --activate` | exit 1, `not signed by any trusted keys` | **exit 0** |
| `corepack prepare pnpm@10.34.5 --activate` | exit 1, same error | **exit 0** |

Both packages in that registry are signed with its own key (`SHA256:5mJAtUoc…`), which is exactly what `<registry>/-/npm/v1/keys` publishes and what Corepack currently has no way to trust:

```bash
curl -s https://npm.cloudsmith.io/opswithranjan/npm/-/npm/v1/keys
curl -s https://npm.cloudsmith.io/opswithranjan/npm/npm/11.16.0 | jq '.dist.signatures'
```

## Notes

- A trailing slash on `COREPACK_NPM_REGISTRY` would have produced `//-/npm/v1/keys`; npmjs tolerates that path but not every registry does, so it's trimmed before the keys path is appended.
- Key expiry is still not enforced, matching existing behaviour (the bundled npm key with `expires: 2025-01-29` is already past that date and remains trusted). Out of scope here.

### DIFF
--- a/README.md
+++ b/README.md
@@ -354,7 +354,11 @@ same major line. Should you need to upgrade to a new major, use an explicit
 
 - `COREPACK_INTEGRITY_KEYS` can be set to an empty string or `0` to
   instruct Corepack to skip integrity checks, or to a JSON string containing
-  custom keys.
+  custom keys. When it is not set and `COREPACK_NPM_REGISTRY` points to a
+  registry other than the default one, Corepack additionally trusts the signing
+  keys that registry publishes at `/-/npm/v1/keys`, so registries which re-sign
+  the packages they serve can still be verified. Those keys are only requested
+  when the bundled npm keys don't already cover the signature.
 
 ## Troubleshooting
 
--- a/sources/corepackUtils.ts
+++ b/sources/corepackUtils.ts
@@ -300,7 +300,7 @@ export async function installVersion(installTarget: string, locator: Locator, {s
       if (signatures! == null || integrity! == null)
         ({signatures, integrity} = (await npmRegistryUtils.fetchTarballURLAndSignature(registry.package, version)));
 
-      npmRegistryUtils.verifySignature({signatures, integrity, packageName: registry.package, version});
+      await npmRegistryUtils.verifySignature({signatures, integrity, packageName: registry.package, version});
       // @ts-expect-error ignore readonly
       build[1] = Buffer.from(integrity.slice(`sha512-`.length), `base64`).toString(`hex`);
     }
--- a/sources/npmRegistryUtils.ts
+++ b/sources/npmRegistryUtils.ts
@@ -4,6 +4,7 @@ import {createVerify}             from 'crypto';
 import defaultConfig              from '../config.json';
 
 import {shouldSkipIntegrityCheck} from './corepackUtils';
+import * as debugUtils            from './debugUtils';
 import * as httpUtils             from './httpUtils';
 
 // load abbreviated metadata as that's all we need for these calls
@@ -13,12 +14,23 @@ export const DEFAULT_HEADERS: Record<string, string> = {
 };
 export const DEFAULT_NPM_REGISTRY_URL = `https://registry.npmjs.org`;
 
-export async function fetchAsJson(packageName: string, version?: string) {
-  const npmRegistryUrl = process.env.COREPACK_NPM_REGISTRY || DEFAULT_NPM_REGISTRY_URL;
+// Standard endpoint at which an npm registry publishes its signing keys.
+// see: https://docs.npmjs.com/about-registry-signatures
+const KEYS_ENDPOINT_PATH = `/-/npm/v1/keys`;
 
-  if (process.env.COREPACK_ENABLE_NETWORK === `0`)
-    throw new UsageError(`Network access disabled by the environment; can't reach npm repository ${npmRegistryUrl}`);
+interface TrustedKey {
+  keyid: string;
+  key: string;
+}
 
+// A configured registry may carry a trailing slash, which would turn the keys
+// path into `//-/npm/v1/keys`; not every registry resolves that to the same
+// route, so it gets trimmed before the path is appended.
+function trimTrailingSlashes(url: string) {
+  return url.replace(/\/+$/, ``);
+}
+
+function getRegistryHeaders() {
   const headers = {...DEFAULT_HEADERS};
 
   if (`COREPACK_NPM_TOKEN` in process.env) {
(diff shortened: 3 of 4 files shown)


opswithranjan:
For context on where this comes from: I work at Cloudsmith, and this reached us through several customers whose builds break against Corepack. It isn't specific to us though. Any registry that re-signs the packages it serves hits the same wall, and the fix is built on the standard `/-/npm/v1/keys` endpoint rather than anything Cloudsmith-specific.

To make this verifiable rather than just described, here is a **public** registry that reproduces it. No credentials needed.

**The setup:** The registry publishes one signing key, and the packages it serves are signed with that key rather than npm's:

```bash
R=https://npm.cloudsmith.io/opswithranjan/npm

# key the registry publishes
curl -s "$R/-/npm/v1/keys" | jq -r '.keys[].keyid'
# -> SHA256:5mJAtUoc0TI7kFjJ8NoV+LnJqHmAsJSv+r7wbMoQrBs

# key that actually signed the package
curl -s "$R/npm/11.16.0" | jq -r '.dist.signatures[].keyid'
# -> SHA256:5mJAtUoc0TI7kFjJ8NoV+LnJqHmAsJSv+r7wbMoQrBs

# npm's keys, for contrast: neither of these is the above
curl -s https://registry.npmjs.org/-/npm/v1/keys | jq -r '.keys[].keyid'
```

**Reproduce the failure on `main`:**

```bash
yarn build
unset COREPACK_INTEGRITY_KEYS
export COREPACK_NPM_REGISTRY=https://npm.cloudsmith.io/opswithranjan/npm
COREPACK_HOME=$(mktemp -d) node dist/corepack.js prepare npm@11.16.0 --activate
```

```
Usage Error: The package was not signed by any trusted keys: {
  "signatures":  [ { "keyid": "SHA256:5mJAtUoc…" } ],
  "trustedKeys": [ "SHA256:jl3bwswu…", "SHA256:DhQ8wR5…" ]
}
```

**Same command on this branch:**

```bash
yarn build
COREPACK_HOME=$(mktemp -d) node dist/corepack.js prepare npm@11.16.0 --activate
```

Exits 0 and activates. Corepack prints nothing beyond the Preparing ... line on success, so a quiet run is the pass. pnpm@10.34.5 in the same registry behaves the same way (fails on main, passes here).

A fresh `COREPACK_HOME` matters on each run, otherwise the cached package manager means nothing is downloaded or verified. `COREPACK_INTEGRITY_KEYS` must be unset, since setting it takes precedence by design and bypasses the code path in question.

Happy to adjust the approach if you'd rather this were shaped differently.

aduh95:
I'm not sure this is desirable, fetching the public keys from HTTPS defeats the purpose. TLS already provides integrity checks, so the only use case for those signature is that you don't trust TLS (maybe you don't trust yourself for keeping your root certificates up-to-date).
That being said, I agree the use-case is valid and we should find a way to address it.

PR #894, opened 2026-09-02
fix: clear last known good cache entry

### LINKED ISSUE #675: `cache clear` does not remove `lastKnownGood.json` from cache
## Current behavior

`corepack cache clear` does not fully clear the Corepack cache. It leaves behind `lastKnownGood.json`, removing only the install folder `v1` from the Corepack cache.

## Expected behavior

The [README](https://github.com/nodejs/corepack/blob/main/README.md) document says;

> `corepack cache clean`
>
> Clears the local `COREPACK_HOME` cache directory.

> COREPACK_HOME can be set in order to define where Corepack should install the package managers. By default it is set to %LOCALAPPDATA%\node\corepack on Windows, and to $HOME/.cache/node/corepack everywhere else.

so the expectation is that after executing `corepack cache clear` on Linux, there will be no files left in the `$HOME/.cache/node/corepack` directory.

Alternatively, if it is intentional to not delete the `lastKnownGood.json` file from the Corepack cache, then the documentation should be changed.

## Steps to reproduce

Ubuntu `24.04.2` LTS, Node.js `22.14.0` LTS, Corepack `0.32.0`

```shell
rm -rf ~/.cache/node/corepack
npm install corepack@latest -g
corepack install -g yarn@latest
corepack cache clear
find ~/.cache/node/corepack
```

## Logs

```text
Installing yarn@4.7.0...
/home/mike/.cache/node/corepack
/home/mike/.cache/node/corepack/lastKnownGood.json
```

wangxpych:
## Summary

- remove `lastKnownGood.json` when running either `corepack cache clean` or `corepack cache clear`
- keep deletion scoped to Corepack-managed cache entries instead of recursively removing the configurable cache root
- cover both command aliases with regression tests

Fixes #675

## Tests

- `yarn build`
- `yarn test --run` (173 passed, 1 expected failure)
- `yarn lint`
- `yarn typecheck`

### DIFF
--- a/sources/commands/Cache.ts
+++ b/sources/commands/Cache.ts
@@ -1,8 +1,9 @@
-import {Command}          from 'clipanion';
-import fs                 from 'fs';
+import {Command}                                 from 'clipanion';
+import fs                                        from 'fs';
+import path                                      from 'path';
 
-import {getInstallFolder} from '../folderUtils.ts';
-import type {Context}     from '../main.ts';
+import {getCorepackHomeFolder, getInstallFolder} from '../folderUtils.ts';
+import type {Context}                            from '../main.ts';
 
 export class CacheCommand extends Command<Context> {
   static paths = [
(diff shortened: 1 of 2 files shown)


wangxpych:
Closing because I’m stepping back from maintaining this contribution. Thanks for your time.

MikeMcC399:
https://agentscan.tools/user/wangxpych

Automation signals

